PR #1, opened 2016-08-23
Binary numbers unit

caitlind:
I've created a basic file structure and skeleton documents for the binary numbers unit and the 3 unplugged lesson plans for the unit. 

Would like to get others opinions on this current layout and file structure, especially Tracy's. Once this is agreed on we can create branches in the same layout for each of the 3 initial units we're developing.

I haven't added documents for the 'Breakdown' and 'Outcomes' lessons yet as I'm not currently sure what the layout and contents for these are.


caitlind:
@timbellnz @tah942

JackMorganNZ:
After talking to original authors, this PR content is out of date and should be closed.

Issue #57, opened 2017-03-03
Restructure requirements files to allow requirements per development environment


JackMorganNZ:
See chapter 5 of Two Scoops of Django 1.8.

PR #63, opened 2017-03-04
Restructure Python requirements files

JackMorganNZ:
This pull request:

- Restructures requirements files for dependencies for specific environments.
- Locks dependencies to specific version to avoid issues introduced by updates.

Will close issues: #51 #57

### DIFF
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -1,0 +1,6 @@
+# Local development dependencies go here
+-r base.txt
+
+# Documentation
+sphinx>=1.5.2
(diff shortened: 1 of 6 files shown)


requirements/local.txt
```diff
@@ -0,0 +1,6 @@
+# Local development dependencies go here
+-r base.txt
+
+# Documentation
+sphinx>=1.5.2
+sphinx_rtd_theme>=0.1.9
```
hayleyavw: Should these be locked?

Issue #74, opened 2017-03-07
Add breadcrumbs


JackMorganNZ:
This will enable easier navigation, especially during development.

JackMorganNZ:
Breadcrumbs were added in: 99623468e5aa0a9784ea7d57d9b0b20a4c6980f4

> I have only added breadcrumbs to ages that
> will be viewed by users. We are still
> unsure how if the resources app can be
> viewed independently, and if programming
> difficulty pages will be viewed, so these
> were skipped.
> 
> The approach of breadcrumbs is that templates
> inherit from the page above them in the
> breadcrumb trail, instead of base.html. This
> adds computational overhead in generating
> templates, but this appears to be very
> minimal.
> 
> It seems like this approach is suitable,
> unless we try to improve template rendering
> speed (but usually other bottlenecks are
> targeted first).

JackMorganNZ:
Basic tests appear to have added 20 to 30 ms of rendering time. Needs further testing.

PR #79, opened 2017-03-08
Add breadcrumbs

JackMorganNZ:
Fixes issue #74. Uses template inheritance to create breadcrumb trail so need to be aware of this when extending templates.


csunplugged/topics/views.py
```diff
@@ -48,16 +48,16 @@ class UnitPlanView(generic.DetailView):
 
     def get_object(self, **kwargs):
         return get_object_or_404(
-            self.model,
+            self.model.objects.select_related(),
```
hayleyavw: what does this do?

Issue #72, opened 2017-03-07
Scratch images aren't watched in default Gulp script


JackMorganNZ:
When `gulp watch` is run, any created Scratch text files are created to images.
When `gulp` is run, which should default to the `watch` task, the images are not created.
The images should be created when `gulp` is run.

Issue #48, opened 2017-03-03
Process Kordac conversion result data


JackMorganNZ:
When a Kordac conversion takes place, the load script should:

- Check images found are within static folder.
- Check glossary entries are defined in the database.
- Generate any required images (scratch images to be added in future Kordac release).

hayleyavw:
@JackMorganNZ you've done the third bullet point here haven't you?

JackMorganNZ:
Yep, three point is done (though with known issue #72).

PR #83, opened 2017-03-09
Render Scratch images (fixes #71)

### LINKED ISSUE #71: Render Scratch images
This requires Kordac `0.3.0` to be released.

JackMorganNZ:
This issue renders Scratch images within Markdown converted by Kordac, using the Gulp script to render Scratch SVGs. There is an issue with Scratch files being watched by the Gulp script, see #72.


csunplugged/utils/BaseLoader.py
```diff
@@ -32,6 +36,29 @@ def setup_md_to_html_converter(self):
...
+        Args:
+            file_path: location of md file to convert
+
+        Returns:
+            Kordac result object
+        """
+        content = open(md_file_path, encoding='UTF-8').read()
+        result = self.converter.convert(content)
+        # Write Scratch data for image rendering by Gulp script
+        if 'scratch_images' in result.required_files:
```
hayleyavw: Maybe this scratch stuff should be in it's own function? `convert_md_file` to me sounds like it's really just converting the text, not manipulating/using the result in anyway

csunplugged/templates/topics/programming_exercise_language_solution.html
```diff
@@ -1,20 +1,20 @@
...
 
+{% load render_html_field %}
 {% load django_bootstrap_breadcrumbs %}
 
 {% block breadcrumbs %}
   {{ block.super }}
   {% breadcrumb "Solution" "" %}
 {% endblock breadcrumbs %}
+
 {% block content %}
```
hayleyavw: should everything inside the `block content` be indented?

JackMorganNZ: Didn't fix as it wasn't the purpose of the issue (want to go through all templates again soon and fix that plus check everything inherits properly. I'll log it as an actual issue now.

csunplugged/utils/BaseLoader.py
JackMorganNZ: Definitely should be in it's own function, was treating it as proof of concept until #48 was completed.

Though the function/s should be called from this function, as we should always run these functions after every Kordac call. It might require a method name change (but I can't think of anything better).

csunplugged/templates/topics/programming_exercise_language_solution.html
JackMorganNZ: Now logged as issue #85

JackMorganNZ:
Justification of PR: The bug doesn't impede the rendering of Scratch files, but limits our ease in using the Gulp script. If the bug didn't render the files, then I wouldn't accept the PR.

csunplugged/utils/BaseLoader.py
JackMorganNZ: Should the Scratch render logic be a function outside the `BaseLoader` class, or a method of it? It doesn't require anything within the loader, so is it a function?

hayleyavw:
okies, happy with that. Will approve once that code is moved to it's own function.

PR #82, opened 2017-03-09
Issue/81

hayleyavw:
Fixes #81

Issue #81, opened 2017-03-09
Always ask for page size and copies for resource generation


hayleyavw:
Add these questions to `base-resource.html` template

PR #84, opened 2017-03-09
Add treasure hunt resource and fixes to resource generation

JackMorganNZ:
Related: #18 #81 #82


csunplugged/resources/views/sorting_network.py
```diff
@@ -49,3 +37,37 @@ def resource_image(get_request, resource):
...
     return image
+
+
+def subtitle(get_request, resource):
+    """Returns the subtitle string of the resource.
+
+    Used after the resource name in the filename, and
+    also on the resource image.
+    """
+    TEMPLATE = '{} to {}'
```
hayleyavw: you're a little inconsistent with naming template strings.
Please make this more descriptive than just "template"

csunplugged/static/scss/website.scss
```diff
@@ -20,3 +20,6 @@ h1, h2, h3 {
     color: #888;
   }
 }
+input[type="text"].long-text-field {
+  width: 20em;
+}
```
hayleyavw: missing newline (from my commit), you might as well fix before merging into develop

csunplugged/templates/resources/treasure-hunt.html
```diff
@@ -0,0 +1,36 @@
+{% extends "resources/base-resource.html" %}
+
+<!-- Name of resource -->
+{% block name %}{{ resource.name }}{% endblock name %}
```
hayleyavw: happy with opening and closing a block on the same line? Or should there be a new line here for consistency? (see lines 7 and 8)

JackMorganNZ: Was to match usage of parent, see base template. We could always write as block as whitespace is irrelevant?

csunplugged/resources/views/sorting_network.py
JackMorganNZ: Suggestions?

JackMorganNZ:
I'm hopeless at remembering to go back and update docstrings after I've written the function. Hopefully I'll learn, but would it be wise to install a style checker to check each function/method has a Google style docstring? That would reduce the mental load of having us to remember?

csunplugged/resources/views/sorting_network.py
JackMorganNZ: `SUBTITLE_TEMPLATE`?

hayleyavw: much better

csunplugged/templates/resources/treasure-hunt.html
hayleyavw: Maybe we should talk about this when you go through all the templates (re your comment on the scratch PR)

hayleyavw:
The problem is probably them becoming out of date quickly rather than not being there at all (in saying that I just found I missed one entirely in kordac :joy:), I think we just need to be more diligent and less lenient on  PR reviews

PR #147, opened 2017-03-20
Issue/77

### LINKED ISSUE #77: Store custom Kordac templates
The custom Kordac templates for Markdown conversion need to be stored within the repository.
Gut instinct is to store these within the `templates` directory under `markdown_templates` and then exclude this folder from the Django template loader (to avoid loading unused templates in serving webpages).

These can then be loaded for Kordac (possibly a Django loader would do the job).

hayleyavw:
Stores html templates for Kordac in separate files.

Fixes #77


docs/source/developer/website_design.rst
```diff
@@ -26,4 +26,13 @@ In summary:
   - For complete examples of the breadcrumb package being used, look at the
     templates for the ``topics`` application.
 
+Setting Custom Converter Templates
+==============================================================================
+We use Kordac to convert Markdown files to HTML. To override a default Kordac
+template, add a new HTML file to ``utils/custom_converter_templates/<processor-name>.html``.
+
+The file name must correspond to the name of a processor in Kordac.
```
JackMorganNZ: This is slightly incorrect, as the majority of templates use a template of the processor name, but some processors use multiple templates. Possibly change to:

> The template file name must correspond to the name of a processor in Kordac (for example: `image.html`, or the name of a supporting template specified in Kordac documentation (for example: `relative-image-link.html`).

Ideally which should just specify base template names in the Kordac docs for every processor at the top, then refer to the Kordac docs from here:

> Processor name: `image`
> Base template: `image.html`

hayleyavw:
ping @JackMorganNZ

PR #156, opened 2017-03-26
Fixed __str__ method for prog ex lang imp model. (Fixes #95)

### LINKED ISSUE #95: Fix __str__ function in ProgrammingExerciseLanguageImplementation
The `__str__` function of `ProgrammingExerciseLanguageImplementation` returns an unknown property which creates an error.

hayleyavw:
Also added language implementation to loader output.


hayleyavw:
Should move the log output for prog ex to be before language.

PR #213, opened 2017-04-25
Issue/183 PEP8 fixes and simplify error string handling

JackMorganNZ:
This fixes all but two PEP8 fixes (so Travis CI will show fail), and changes the ways strings are managed in custom exceptions.

@hayleyavw: The Learning Outcomes loader is broken, the wrong variables are used in the `__init__` method so rather than fixing it, I'll leave it to you (as the rest of it might need changing too). We haven't encountered this yet as we don't use it.

I also haven't gotten around to testing the output of these errors, I'll get to that later tonight.

### DIFF
--- a/csunplugged/utils/errors/CouldNotFindConfigFileError.py
+++ b/csunplugged/utils/errors/CouldNotFindConfigFileError.py
@@ -1,5 +1,7 @@
 from .Error import Error
 
+ERROR_MESSAGE = "\nCould not find config file.\n"
+
 
 class CouldNotFindConfigFileError(Error):
     '''Raised when a yaml file cannot be found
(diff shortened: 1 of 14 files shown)


JackMorganNZ:
I've checked all the error messages and fixed a broken one. I also added some more text to messages where the error wasn't stated in English (only in exception name).

I noticed that the following errors are never used and had a odd bug:

- `MissingCurriculumArea`
- `MissingLanguageImplementation`
- `MissingLearningObjective`
- `MissingRequiredFieldError`

These errors asked for a `loader` instead of a `filename`, which I think is incorrect (as you would give it the place the error occurred. However I don't know if that is what you intended as they are never used in the loaders.

I changed all the errors to `filename` instead of `loader`, but change back if needed.

The issues from the initial pull request text still exist.

PR #239, opened 2017-04-29
Documentation updates (fixes #129 #204 #238)

### LINKED ISSUE #129: Add guide for authors about issue management and Git branching
## Description

A guide should be explaining to authors (for all content) how to:

- Create issues for work to be completed
- When to assign workers to issues
- When and why to branch with Git

### Checklist

*Change the space in the box to an `x` for those you have completed. You can also fill these out after creating the issue.*

- [x] I have linked any relevant existing issues/suggestions in the description above (include `#???` in your description to reference an issue, where `???` is the issue number).

### LINKED ISSUE #204: Update documentation flowcharts with correct commands
Flowcharts stating to run `updatedata` need to be updated to new commands.

Should be started after #188 is merged in.

Created from work in #193.

### LINKED ISSUE #238: Tidy 'System Commands for Developing' documentation
- Provide clearer filename.
- Remove accidental sentence.

JackMorganNZ:
## Proposed changes

- Add extra project style details for authors (fixes #129)
- Update documentation flowcharts with correct commands (fixes #204)
- Tidy helper commands documentation (fixes #238)
- Update troubleshooting documentation to use `csu` commands.


docs/source/author/topics.rst
```diff
@@ -41,19 +41,16 @@ diagram.
...
-
-.. _viewing-the-current-state:
-
-Viewing the Current State
+Viewing All Topics Content
 ------------------------------------------------------------------------------
 
-When developing locally, once you have a server and ``gulp`` running (see
-:doc:`../getting_started/basic_usage`) you can go to the url below to get a
+When developing locally, once you have run ``csu start`` (see
```
hayleyavw: there is quite a bit of inconsistency here. We run `./csu start` but throughout the whole docs we don't prefix commands with the `./`, i.e. `csu` - this doesn't even work on my machine, I have to have the `./` there, so I think we should include this in all the docs rather than just mention it once at the start, what do you think?

docs/source/author/topics.rst
```diff
@@ -81,81 +78,130 @@ The following flow charts will take you step by step through the process of addi
...
+.. The image is included as raw HTML because it has clickable nodes.
+.. raw:: html
+
+  <map name="programming-exercises-map">
+    <area shape="rect" coords="554,470,633,505" href="#programming-exercises-configuration-file">
+    <area shape="rect" coords="240,572,317,605" href="#programming-exercises-configuration-file">
+    <area shape="rect" coords="240,690,317,725" href="#programming-exercises-structure-configuration-file">
+    <area shape="rect" coords="240,807,317,841" href="../getting_started/helper_commands.html#update">
+  </map>
+  <img src="../_static/img/topics_adding_programming_exercises_flowchart.png" usemap="#programming-exercises-map">
```
hayleyavw: this diagram missed the step to add the prog ex config file to the topic config file :(

docs/source/getting_started/helper_commands.rst
```diff
@@ -40,6 +38,21 @@ Where ``[COMMAND]`` is a word from the list below:
...
 
+All users of the project (content and technical developers) should become
+familiar with the following commands:
+
+- :ref:`start`
+- :ref:`end`
+- :ref:`build`
+- :ref:`update`
+
+Technical developers should also understand the following:
```
hayleyavw: "should also understand *at least* the following"

docs/source/troubleshooting.rst
```diff
@@ -24,13 +24,13 @@ Viewing website
 Images are not displayed when I view the website
```
hayleyavw: well, if you are talking about embedded images or icons, the converter will spaz and raise an exception all by itself and tell you to check the file exists/is named correctly/is in the right place, does this need to be in the faq?...

docs/source/author/topics.rst
JackMorganNZ: I was unsure where to strike the balance of putting that in, because it really comes down to the user understanding their operating system. I only put it once because I know it would work in that situation as the user is in the correct working directory. From that point onwards, I can't be sure that they are in the right place. Should we just assume that and always put `./` before? Or should we assume understanding of how to run a file? Thoughts?

hayleyavw: well, what if we put `./` everywhere, and then put a note in the troubleshooting page that if this doesn't work, it means they are in the wrong place?

JackMorganNZ: Sounds good!

JackMorganNZ:
@hayleyavw This is ready for re-reviewing.

PR #254, opened 2017-05-01
Add commands for deleting unused volumes, containers, and images

JackMorganNZ:
Also stops creating new images for each `run` command.


JackMorganNZ:
Yes, `--rm` deletes the old containers as they go.

Issue #259, opened 2017-05-03
Load glossary keys before glossary content


JackMorganNZ:
Currently as each glossary key is discovered, the content is processed. If the glossary definition links to another glossary term that hasn't been loaded yet, this raises an exception. While the occurrence of this issue would be lowered by #43, it could still occur. It could also be lowered by the order terms are defined in, but circular definitions would cause this issue.

A simpler solution would be to do the following process:

- For each key, check the content file exists, then create/update object without term/content.
- For each key/object again, load term/content.

PR #263, opened 2017-05-03
Add glossary (fixes #119)

### LINKED ISSUE #119: Add teacher glossary
## Description

Add site wide glossary for teachers (see #101).

### Checklist

*Change the space in the box to an `x` for those you have completed. You can also fill these out after creating the issue.*

- [x] I have linked any relevant existing issues/suggestions in the description above (include `#???` in your description to reference an issue, where `???` is the issue number).

JackMorganNZ:
This pull request adds the following:

- Glossary model for storing definitions.
- Definitions are loaded from configuration files (see new documentation).
- Basic test cases for glossary url and views (loaders are currently not tested as base loader tests are yet to be written).
- Display link to glossary entry in text, and pop up definition on click.
- Glossary page with all term/definitions (currently not linked to by any page).

This pull request also updates the `csu` script to separate 'standard' and 'dev' commands.

Issues created from the work of the pull request:
- #259 Load glossary keys before glossary content.
- Allow glossary to work in different languages.
- uccser/verto#190 - Glossary term not saved if back reference not given.
  - Once this issue is completed, the checking of glossary terms will need to be tested.

Placeholder content should eventually be deleted.

Fixes #119.


JackMorganNZ:
- One of the Travis CI builds failed due to conflicting migration files? Rebuilding it now to check.
- I want to test an image in a glossary definition this morning.

csunplugged/static/js/website.js
```diff
@@ -0,0 +1,41 @@
...
+  // Find glossary modal
+  var glossary_modal = $("#glossary-modal");
+
+  // If glossary modal already has content for term
+  if (glossary_modal.attr("data-glossary-term") == slug) {
+    glossary_modal.modal('show');
+  }
+  // Otherwise download HTML of modal and add to DOM
+  else {
+    // TODO: Allow URL to work for different languages
```
hayleyavw: should this be transferred to an issue rather than a TODO in the code?

csunplugged/static/js/website.js
```diff
@@ -0,0 +1,41 @@
...
+      url: url,
+      data: "term=" + slug,
+      async: true,
+      cache: true,
+      dataType: "json",
+      success: update_glossary_modal
+    });
+  }
+}
+
```
hayleyavw: lines 1-31 are quite heavily commented, this isn't really necessary unless it's mathematical/extremely specific code, the code should speak for itself, just a docstring is enough (feel free to argue with me)

csunplugged/static/js/website.js
```diff
@@ -0,0 +1,41 @@
...
+  var slug = $(this).data("glossary-term");
+  // Find glossary modal
+  var glossary_modal = $("#glossary-modal");
+
+  // If glossary modal already has content for term
+  if (glossary_modal.attr("data-glossary-term") == slug) {
+    glossary_modal.modal('show');
+  }
+  // Otherwise download HTML of modal and add to DOM
+  else {
```
hayleyavw: this should be on the same line as the closing parenthesis on 15

csunplugged/templates/dev/index.html
```diff
@@ -80,6 +81,19 @@
...
+    <ul>
+      {% for glossary_term in glossary_terms %}
+        <li>
+          <h3>{{ glossary_term.term }}</h3>
+          <p>Key: <code>{{ glossary_term.slug }}</code></p>
+          {% render_html_field glossary_term.definition %}
+        </li>
+      {% endfor %}
+    </ul>
+  {% endif %}
```
hayleyavw: I am slightly concerned that this is going to cause the dev page to get way too long, can the layout of this please be compacted down somehow (not a big title for each word maybe? just bolded text?) and put at the bottom of the page.

csunplugged/tests/topics/urls/test_glossary_url.py
```diff
@@ -0,0 +1,13 @@
...
+
+class GlossaryURLTest(BaseTestWithDB):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.language = 'en'
+
+    def test_valid_glossary_url(self):
+        url = reverse('topics:glossary')
+        self.assertEqual(url, '/en/topics/glossary/')
```
hayleyavw: shouldn't this also be testing for the correct response? (404, 200?)

hayleyavw: should be using double quotes, not single

csunplugged/topics/management/commands/_GlossaryTermsLoader.py
```diff
@@ -0,0 +1,52 @@
...
+
+from utils.BaseLoader import BaseLoader
+from topics.models import GlossaryTerm
+
+
+class GlossaryTermsLoader(BaseLoader):
+    """Custom loader for loading glossary terms."""
+
+    def __init__(self, glossary_folder_path, glossary_terms, structure_file_path, BASE_PATH):
+        """Create the loader for loading programming exercises.
```
hayleyavw: uhh...copy/pasted docstring?... caught ya!

csunplugged/topics/management/commands/_GlossaryTermsLoader.py
```diff
@@ -0,0 +1,52 @@
...
+            filename = "{term}.md".format(term=glossary_slug)
+            definition_file_path = os.path.join(
+                self.BASE_PATH,
+                filename
+            )
+
+            glossary_term_content = self.convert_md_file(
+                definition_file_path,
+                self.structure_file_path
+            )
```
hayleyavw: there should be error handling there - what happens if it can't find the file? is missing a header? is missing the definition?

csunplugged/topics/management/commands/loadtopics.py
```diff
@@ -60,6 +61,23 @@ def handle(self, *args, **options):
...
+            else:
+                raise MissingRequiredFieldError(
+                    structure_file_path,
+                    ["glossary-folder"],
+                    "Application Structure"
+                )
+            if glossary_folder_path is not None:
+                GlossaryTermsLoader(
+                    glossary_folder_path,
+                    structure_file["glossary-terms"],
```
hayleyavw: I don't think this is really necessary. I think that if we ask for the name of the folder, then we just read every md file from there as a glossary term, no need to ever list them in the structure file as we can use the file name as the key/slug.

csunplugged/topics/management/commands/_GlossaryTermsLoader.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        super().__init__(BASE_PATH)
+        self.glossary_terms = glossary_terms
+        self.structure_file_path = structure_file_path
+        self.BASE_PATH = os.path.join(self.BASE_PATH, glossary_folder_path)
+
+    @transaction.atomic
+    def load(self):
+        """Load the glossary content into the database."""
+        for glossary_slug in self.glossary_terms:
```
hayleyavw: this is going to break if `self.glossary_terms` is none (which is a potential case and needs to be handled)

csunplugged/topics/views.py
```diff
@@ -337,3 +340,36 @@ class OtherResourcesView(generic.DetailView):
...
+            GlossaryTerm,
+            slug=glossary_slug
+        )
+        data = {
+            "slug": glossary_slug,
+            "term": glossary_item.term,
+            "definition": render_html_with_static(glossary_item.definition)
+        }
+        return JsonResponse(data)
+    # Otherwise return glossary webpage with all terms
```
hayleyavw: so...if I request a term that doesn't exist, then I'm taken to the glossary page (when I'm expecting just a modal to pop up) ? Is that correct? That seems like odd behaviour.

csunplugged/utils/BaseLoader.py
```diff
@@ -58,7 +59,8 @@ def convert_md_file(self, md_file_path, config_file_path, heading_required=True)
         """Return the Verto object for a given Markdown file.
 
         Args:
-            md_file_path: location of Markdown file to convert
+            md_file_path: Location of Markdown file to convert
+            config_file_path: Path to related the config file, used for errors.
```
hayleyavw: seeing as you are fixing these, you might as well specify that these are strings too

csu
```diff
@@ -52,6 +55,10 @@ function start {
   # Alert user that system is ready
   printf "\n${GREEN}Systems are ready!${NC}\n"
 }
+function start_help {
+  printf "   start             Start development environment (this also runs\n"
```
JackMorganNZ: No padding arguments that I could find.

csu
```diff
@@ -155,92 +220,125 @@ function clean {
...
+        exit
+        ;;
+      shell)
+        shell
+        exit
+        ;;
+      static)
+        static
+        static_scratch
+        collect_static
```
JackMorganNZ: Was more for future proofing, so that command calls three separate functions. If we wanted to add `dev scratch` later, then it would just call the last two.

csunplugged/static/js/website.js
JackMorganNZ: Both, in the past (for example: resource generation) I create TODOs when developing to find the problem area easier, and also as a reminder to create the issue. Will list it in the PR.

csunplugged/tests/topics/urls/test_glossary_url.py
JackMorganNZ: No, as the response is calculated by a view so it's in the view tests. The URL tests only check the URLs calculate correctly.

I did notice the existing tests were checking status in both the URL and view tests, but it made sense for it only to be tested in the views.

csunplugged/static/js/website.js
JackMorganNZ: Changed but not shown in these lines.

JackMorganNZ:
@hayleyavw Ready for review again.

csunplugged/topics/management/commands/_GlossaryTermsLoader.py
```diff
@@ -0,0 +1,50 @@
...
+        self.BASE_PATH = os.path.join(self.BASE_PATH, glossary_folder_path)
+
+    @transaction.atomic
+    def load(self):
+        """Load the glossary content into the database."""
+        for filename in listdir(self.BASE_PATH):
+            glossary_file_name = re.search(r"(.*?).md$", filename)
+            if glossary_file_name:
+                glossary_slug = glossary_file_name.groups()[0]
+                glossary_file_path = os.path.join(self.BASE_PATH, filename)
```
hayleyavw: usually split this over multiple lines

csunplugged/templates/dev/index.html
```diff
@@ -80,6 +81,13 @@
     </ul>
   {% endif %}
 
+  {% if glossary_terms %}
```
hayleyavw: delete me

PR #285, opened 2017-05-05
Improve curriculum integration badges to show parent (fixes #276)

### LINKED ISSUE #276: Format curriculum integration badges to show parent
When listing a curriculum integration that has a parent, the parent should be listed. 

For example, it could be rendered as:

```
(English: Writing) (Science: Physics) (Arts)
```

These should also act as a link to a search of all related material to the area.

JackMorganNZ:
This pull request does the following:

- Badges now show parent prefix if parent exists, retrieved from method of model.
- Colours groups of curriculum areas.
- Use template for badge creation to remove template duplication.

Fixes #276.


csunplugged/topics/management/commands/_CurriculumAreasLoader.py
```diff
@@ -56,10 +56,19 @@ def load(self):
                     "Curriculum Area"
                 )
 
+            curriculum_area_colour = curriculum_area_data.get("colour", None)
+            if curriculum_area_colour is None:
+                raise MissingRequiredFieldError(
+                    self.structure_file_path,
+                    ["colour"],
```
hayleyavw: check MissingRequiredFieldError.py, you actually have to specify *all* of the required fields, not just the one that is missing as this is the error message it gives:

> A {model} requires the following fields:
{fields}
One or more of these fields are missing.

csunplugged/topics/models.py
```diff
@@ -45,6 +45,7 @@ class CurriculumArea(models.Model):
     #  Auto-incrementing 'id' field is automatically set by Django
     slug = models.SlugField(unique=True)
     name = models.CharField(max_length=100, unique=True)
+    colour = models.CharField(max_length=15, null=True)
```
hayleyavw: this should be longer. What if we eventually just use web colour codes rather than mapping the colours ourselves? https://en.wikipedia.org/wiki/Web_colors - several of those colours are longer than 15 characters long.

csunplugged/topics/models.py
```diff
@@ -18,7 +18,7 @@ def __str__(self):
         """Text representation of GlossaryTerm object.
 
         Returns:
-            Term attribute of GlossaryTerm (string).
+            Term attribute of GlossaryTerm (str).
```
hayleyavw: why are all of these changed to `str` ? All the loaders use `string`...

docs/source/author/topics.rst
```diff
@@ -613,11 +613,14 @@ Curriculum Areas Configuration File
 
       - ``name:`` The name of the curriculum area (this is what will be displayed to the
         user).
+      - ``colour:`` The CSS colour class to use for colouring the curriculum
+        area badge on the website.
+        This colour is also applied to all children of curriculum area.
```
hayleyavw: didn't we define all our own colours, how does an author see the list of available colours when picking the colour to use in the config file?

csunplugged/topics/management/commands/_CurriculumAreasLoader.py
JackMorganNZ: Why don't we switch it to a specific field? Wouldn't that be easier for authors?

csunplugged/topics/models.py
JackMorganNZ: This is because it's only a CSS class name, 15 characters seemed generous enough for a class name but can expand it if you think it would be a problem.

csunplugged/topics/models.py
JackMorganNZ: Since we were using `dict` for dictionary, it seemed appropriate to use `str` for string. However specifying the type is only required in the description by the style guide, what would you like to do?

- Full text of type in brackets?
- Python notation of type in brackets?
- Mention type in description?
- Something else?

csunplugged/topics/models.py
JackMorganNZ: This is because it's only a CSS class name, 15 characters seemed generous enough for a class name but can expand it if you think it would be a problem.

docs/source/author/topics.rst
JackMorganNZ: This is a bit of a chicken and egg situation, so will do my best to explain it.

Originally I didn't have the colour specified, and had CSS classes written to pick out based off the area's `slug`, however this would fail with different translations.

Next option was to let the author specify the colour in the configuration file using a hexadecimal code, or something similar, however you would end up with colours that don't match the rest of the website's branding. I didn't want to the responsibility of web design on the authors.

So I chose to specify a CSS class, that is defined in the SCSS files and uses colours from our branding colour palette. When translated, these values should be left how they are, similar to numerical values for lessons. A list of available colours could be defined in the documentation under this configuration file? I only added enough colours to cover all these groups (but don't know how final this list is), so eventually I should add probably twice as many.

I couldn't figure out a better solution, but this seemed to work so far.

csunplugged/topics/management/commands/_CurriculumAreasLoader.py
hayleyavw: we could and probably should, when I first started implementing it I did it this way so as to not get duplicate code everywhere, but can probably find a way to minimise that now. Maybe use the full list here and make an issue to go and change that error (and all calls to it) later?

csunplugged/topics/models.py
hayleyavw: type in *parenthesis* is fine, might as well do python notation because it is shorter and we know what it means.

fun fact, a "bracket" is actually this: `{`

docs/source/author/topics.rst
hayleyavw: I think it should just be added to the documentation, just a note somewhere listing all the available colours so that people don't have to go digging through the source code.

csunplugged/topics/management/commands/_CurriculumAreasLoader.py
JackMorganNZ: Sounds good, will do.

csunplugged/topics/models.py
JackMorganNZ: Mind blown...

Also sounds good.

docs/source/author/topics.rst
JackMorganNZ: Will do!

PR #297, opened 2017-05-07
Load glossary keys before glossary content (fixes #259)

### LINKED ISSUE #259: Load glossary keys before glossary content

JackMorganNZ:
This pull request now creates all the glossary objects with `slug` values, then reiterates over the objects to add `term` and `content` values. This is to avoid the issue of a glossary entry linking to an undefined term during loading.

**Note:** The pull request is failing as the number of misses in code coverage increases due to new code added to loaders.


csunplugged/topics/management/commands/_GlossaryTermsLoader.py
```diff
@@ -23,29 +22,33 @@ def __init__(self, glossary_folder_path, structure_file_path, BASE_PATH):
...
+                    slug=glossary_slug,
+                )
+                glossary_term.save()
+
+        for glossary_slug in glossary_slugs:
+                glossary_term = GlossaryTerm.objects.get(slug=glossary_slug)
                 glossary_file_path = os.path.join(
                     self.BASE_PATH,
-                    filename
+                    glossary_slug + self.FILE_EXTENSION
```
hayleyavw: much prefer `.format()` rather than concatenation, more consistent across the rest of the loaders

csunplugged/topics/models.py
```diff
@@ -11,7 +11,7 @@ class GlossaryTerm(models.Model):
 
     #  Auto-incrementing 'id' field is automatically set by Django
     slug = models.SlugField(unique=True)
-    term = models.CharField(max_length=200, unique=True)
+    term = models.CharField(max_length=200)
```
hayleyavw: why shouldn't a term be unique? What if there is an `algorithm.md` with heading `Algorithm` but I don't notice and create a `algorithms.md` with heading `Algorithm`. Now I have identical terms with different slugs which seems silly.

JackMorganNZ: I agree, but because of the new creation technique, having `unique=True` fails in the first phase of creation as there are multiple `GlossaryTerm` objects with the term `""`.

However would having `term = models.CharField(max_length=200, unique=True, null=True)` solve this? I could test that out?

JackMorganNZ: Worked perfectly! Doesn't allow duplicates (two different files with the same title) but they can be created as `null` to begin with. Ready for re-review.

PR #302, opened 2017-05-08
Fix Django Debug Toolbar to render in local development environment

### LINKED ISSUE #201: Django Debug Toolbar doesn't display in local Docker development environment
Created from work in #193.

JackMorganNZ:
Fixes #201.

### DIFF
--- a/csunplugged/config/settings/testing.py
+++ b/csunplugged/config/settings/testing.py
@@ -4,6 +4,14 @@
 from .base import *  # noqa: F403
 
 
+# DATABASE CONFIGURATION
+# ----------------------------------------------------------------------------
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
+DATABASES = {
+    'default': env.db('DATABASE_URL'),  # noqa: F405
+}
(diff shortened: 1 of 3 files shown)


csunplugged/config/settings/testing.py
```diff
@@ -14,7 +22,7 @@
 # ----------------------------------------------------------------------------
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#secret-key
 # Note: This key only used for development and testing.
-SECRET_KEY = env('DJANGO_SECRET_KEY', default='CHANGEME!!!')  # noqa: F405
+SECRET_KEY = env('DJANGO_SECRET_KEY', default='l@@)w&&%&u37+sjz^lsx^+29y_333oid3ygxzucar^8o(axo*f')  # noqa: F405
```
ravenmaster001: Should this be here?

JackMorganNZ: This is the same value used for local development, so thought I would make them consistent. The production key is still hidden.

PR #315, opened 2017-05-13
Automate deployment of development system (fixes #310)

### LINKED ISSUE #310: Automate deployment to development servers
To investigate:

- [x] ~Setting `gsutil` to do complete sync.~ Will not use for the moment, see: https://cloud.google.com/storage/docs/gsutil/commands/rsync#be-careful-when-using--d-option
- [x] Using script to set environment variables for the specific deployment.
- [x] Check for secrets within logs.
- [x] Use new database account credentials.
  - [x] Delete old credentials once deployed.
- [x] Set `.travis.yml` to deploy on `develop`.
- [x] Update documentation.

JackMorganNZ:
## Proposed changes

This pull request automatically deploys the system to the development website for each commit on the `develop` branch. This is done by a script on Travis CI, the script contains documentation about each step required. Sensitive data is archived together and encrypted by Travis, the raw files are found in our password management system.

This pull request also:
- Renames `compose` folder to `infrastructure` for more general infrastructure use.
- Uses a new service account for deploying development, old account is now deleted.
- Removes programming exercises from lessons two and three of Binary Numbers (as discussed in the last stakeholder meeting). I did this in this pull request to easily test the system had updated the database as required.

*Note:* The `gsutil rsync` command for updating static files is often used with the `-d` parameter to do a complete sync (which means deleting files that don't exist on source). For the moment, this is omitted and copies across new files and leaves old ones there. This is due to the `render` system wanting to store static files on the same bucket (could put on different bucket later).

All logs for commits on Travis (except last few) have been deleted due to connection name of database being displayed (though I still can't find if this should be kept hidden).

The failing tests are due to code coverage, as we only test with the `testing` settings file, so specific lines are never hit (due to `DEBUG=False`). It might be worth ignoring these lines in coverage, or creating new tests to hit these.

Fixes #310.


.travis.yml
```diff
@@ -1,40 +1,40 @@
...
 env:
-  - DOCKER_COMPOSE_VERSION=1.11.2
-
+- DOCKER_COMPOSE_VERSION=1.11.2
 language: python
-
 python:
-  - "3.4"
-
+- '3.6'
```
ravenmaster001: Is this for venv?

.travis.yml
```diff
@@ -1,40 +1,40 @@
...
-  - chmod +x docker-compose
-  - sudo mv docker-compose /usr/local/bin
-  - sudo docker --version
-  - sudo docker-compose --version
-
+- sudo apt-get update
+- sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
+- sudo rm /usr/local/bin/docker-compose
+- curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname
+  -s`-`uname -m` > docker-compose
```
ravenmaster001: Is this still part of the previous line? It seems odd and wrong that this is broken over multiple lines.

.travis.yml
```diff
@@ -1,40 +1,40 @@
...
-  - ./csu dev docs
-
+- "./csu dev test"
+- "./csu dev test_backwards"
+- "./csu dev test_coverage"
+- "./csu dev style"
+- "./csu dev docs"
+cache:
+  directories:
+    - ${HOME}/google-cloud-sdk
```
ravenmaster001: Is there anything sensitive within this cache?

csunplugged/topics/content/en/binary-numbers/unit-plan/unit-plan.yaml
```diff
@@ -31,47 +31,9 @@ reinforcing-sequencing-in-binary-number-systems:
   max-age: 10
   number: 2
   duration: 60
-  programming-exercises:
```
ravenmaster001: Should this be in this PR?

csunplugged/topics/content/en/binary-numbers/unit-plan/unit-plan.yaml
```diff
@@ -31,47 +31,9 @@ reinforcing-sequencing-in-binary-number-systems:
...
-    - number-bits-representing-number
-    - binary-cards-representing-number
-    - decimal-to-binary
 
 codes-for-letters-using-binary-representation:
   min-age: 8
   max-age: 10
   number: 3
   duration: 45
-  programming-exercises:
```
ravenmaster001: REPEAT: Should this be in this PR?

.travis.yml
JackMorganNZ: Possibly, from this documentation it references a file with sensitive info (but not where it is). Let's play it safe and not cache and rebuilt from scratch each time. 
https://cloud.google.com/storage/docs/gsutil/addlhelp/SecurityandPrivacyConsiderations

I had added it because sometimes the SDK was briefly unavailable for download.

.travis.yml
JackMorganNZ: It is from the previous line, but the Travis CI YAML styler did this. It also does this:

```
key:
- value
```

instead of 

```
key:
  - value
```

which I find odd.

I'll edit that line to be one long one instead.

.travis.yml
JackMorganNZ: Yes, see: https://docs.travis-ci.com/user/languages/python/

JackMorganNZ:
Thanks @ravenmaster001, will wait for both reviews from @hayleyavw and @jordangriffiths01.

.travis.yml
```diff
@@ -1,40 +1,36 @@
...
-  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
-  - chmod +x docker-compose
-  - sudo mv docker-compose /usr/local/bin
-  - sudo docker --version
-  - sudo docker-compose --version
-
+- sudo apt-get update
+- sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
+- sudo rm /usr/local/bin/docker-compose
+- curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
```
jordangriffiths01: Always quote strings with variable expansion, also $() instead of backticks for command substitution.

eg: curl -L "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)" > docker-compose

csunplugged/config/settings/database_proxy.py
```diff
@@ -0,0 +1,22 @@
...
+# DATABASE CONFIGURATION
+# ----------------------------------------------------------------------------
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.postgresql",
+        "HOST": "127.0.0.1",
+        "PORT": "5433",
+        "NAME": "csunplugged",
+        "USER": env("GOOGLE_CLOUD_SQL_DATABASE_USERNAME"),  # noqa: F405
```
jordangriffiths01: When/how do these env vars get populated on travis?

JackMorganNZ: These are loaded by `. ./load-develop-deploy-envs.sh` in the `develop-deploy.sh`. The script is a secret file that looks something like this:

```
#!/bin/bash
# Used when running local Django for updating development database.

export DJANGO_SECRET_KEY="SECRET VALUE!"
export GOOGLE_CLOUD_SQL_CONNECTION_NAME="SECRET VALUE!"
export GOOGLE_CLOUD_SQL_DATABASE_USERNAME="SECRET VALUE!"
export GOOGLE_CLOUD_SQL_DATABASE_PASSWORD="SECRET VALUE!"
```

PR #323, opened 2017-05-15
Add binary cards resource

JackMorganNZ:
## Proposed changes

This pull request adds the Binary Cards resource, with the following configuration options:

- Numbers, on or off.
- Black backing, on or off.

This also removes the number of copies option as a default for all forms, but can be included using a template tag. This is because this resource should only ever generate one copy at a time.


csunplugged/tests/resources/ResourcesTestDataGenerator.py
```diff
@@ -0,0 +1,29 @@
...
+    """Class for generating test data for resource tests."""
+
+    def create_test_resource(self, slug, name, webpage_template, generation_view):
+        """Create resource object.
+
+        Args:
+            slug: String of resource slug.
+            name: String of resource name.
+            webpage_template: String describing path to HTML form template.
+            generation_view: String of Python view module filename.
```
hayleyavw: (maybe I'm wrong) but didn't we decided to give the type using python types? e.g. `resource slug (str).` ?

csunplugged/topics/content/en/binary-numbers/unit-plan/unit-plan.yaml
```diff
@@ -26,8 +26,8 @@ how-binary-digits-work:
     - binary-correct-representation
     - no-physical-zeros-ones
   generated-resources:
-    sorting-network:
-      description: One per student.
+    binary-cards:
+      description: One set.
```
hayleyavw: does it make sense to move the description into the `resources.yaml` Or will there be a chance that one resource could have different descriptions?

JackMorganNZ: Definitely should remain in the YAML file, as it's likely different lessons use an interactive differently.

For example:
- One per student.
- One per group.

Issue #17, opened 2017-02-24
Add documentation on resources application


JackMorganNZ:

JackMorganNZ:
This has been partially completed in PR #556.

JackMorganNZ:
The [documentation](http://cs-unplugged.readthedocs.io/en/latest/developer/resources.html) appears to be at a decent level now that I'll close this issue.

Issue #322, opened 2017-05-15
Switch Docker `run` commands to `exec`


JackMorganNZ:

ravenmaster001:
Investigate the `-T` option and if that is necessary too.

PR #339, opened 2017-05-18
Provide static PDF files until render system is deployed (fixes #338)

### LINKED ISSUE #338: Provide static PDF files until render system is deployed
- [x] Add `makeresources` command.
- [x] Add redirect to files in production.
- [x] Add copies attribute to `Resource` model.
- [x] Set resources with copies to create 30 pages.
- [x] Update interface in production.
- [x] Add documentation.

JackMorganNZ:
This pull request adds the following command:

- Add `makeresources` command. This creates all combinations of resources as static PDFs.
- Adds redirection to static files in production environment.
- Add copies attribute to `Resource` model for easy detection if copies are allowed.
- Resources with copies are generated with 30 pages by the `makeresources` command.
- Update interface to show 'Download' in production, and 'Generate' in development.
- Runs both `updatedata` and `makeresources` as smoke tests.
- Add new `flush`, `makeresources`, and `test_specific` commands to `csu` script.

Currently there is no documentation for resources, so I've labelled #17 as 'urgent' to work on it next.

Notes:

- Currently a few resources are generated multiple times (for example, 'no numbers sorted' and 'no numbers unsorted' are technically two combinations on the matrix, but produce the same file. It's not possible to check if the filename exists before creation as this occurs during creation. It could be possible to refactor when the filename creation occurs, however this seems appropriate for a new issue as it's a very small gain for a lot of work.
- The `makeresources` command is called twice, once during testing, and then again during deployment. It could just be called once, but it seems inappropriate to depending on a test case output.


csu
```diff
@@ -91,6 +94,14 @@ function update_help {
   printf "                     and build static files.\n"
 }
 
+# Run Django flush command
+function flush {
+  docker-compose run --rm django /docker_venv/bin/python3 ./manage.py flush
```
JackMorganNZ: Will be completed in #322.

PR #344, opened 2017-05-20
Update LICENSE

### LINKED ISSUE #22: Update LICENSE

JackMorganNZ:
This pull request adds license files for the core system (MIT), the content (CC), and third party licenses.

Fixes #22.
Also closed the 'Ready for Authors' milestone.


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
ravenmaster001: I would like something here to explain more explicitly what this applies too. (Even if it redirects you to the documentation.)

LICENSE-CONTENT
```diff
@@ -0,0 +1,427 @@
+Attribution-ShareAlike 4.0 International
```
ravenmaster001: I would like something here to explain more explicitly what this applies too. (Even if it redirects you to the documentation.)

JackMorganNZ:
Thanks for the suggestion @ravenmaster001. I've clarified the coverage of licenses in the `README`, and made a note at the top of the `LICENSE-CONTENT` file. I've left the `LICENSE` file as is, because I want it easily identifiable by license bots (having a `LICENSE` file with extra content is placed as unknown).

PR #468, opened 2017-06-18
True/False rather than Yes/No in resource code (fixes #342)

### LINKED ISSUE #342: Use true/false instead of yes/no for resource parameter values
These should be interpreted as boolean values by the `retrieve_query_parameter` function.

hayleyavw:


csunplugged/resources/views/binary_cards.py
```diff
@@ -34,7 +34,7 @@ def resource_image(request, resource):
     display_numbers = retrieve_query_parameter(request, "display_numbers", parameter_options["display_numbers"])
     black_back = retrieve_query_parameter(request, "black_back", parameter_options["black_back"])
 
-    if display_numbers == "yes":
+    if display_numbers is True:
```
JackMorganNZ: I would expect this to use the Python convention of `if display_numbers:`

csunplugged/tests/resources/views/test_binary_cards_small.py
```diff
@@ -45,15 +45,24 @@ def test_binary_cards_small_resource_generation_valid_configurations(self):
         combinations = [dict(zip(valid_option_keys, product)) for product in itertools.product(*(valid_options[valid_option_key] for valid_option_key in valid_option_keys))]  # noqa: E501
         print()
         for combination in combinations:
+            print(combination)
```
JackMorganNZ: Debugging statement

csunplugged/tests/resources/views/test_binary_cards_small.py
```diff
@@ -45,15 +45,24 @@ def test_binary_cards_small_resource_generation_valid_configurations(self):
...
             print("   - Testing combination: {} ... ".format(combination), end="")
-            url = base_url + query_string(combination)
+            url_combination = {}
+            for parameter in combination:
```
JackMorganNZ: Optional: It might be better to replace this code with the `util` function now to save this duplication.

PR #484, opened 2017-06-19
Add initial content for unplugged programming topic

hayleyavw:
~~**NOTE:** this PR **should not be merged** unless there is at least one lesson for each of the three unit plans (Kidbots, Geometry and Numeracy) otherwise place holder lessons will go to develop. The place holder lessons could be removed, but this will cause the loaders to throw an error, therefore interfering with adding content in other areas.~~

When reviewing this PR:
- [x] The Unit Plans need **very** careful review as they are very similar so I need to make sure I didn't copy something and forget to change it to the correct wording (e.g. "kidbots" rather than "numeracy")
- [x] The bullet points in Kidbots lesson 1 (5-7) render with inconsistent spacing, if you could suggest a fix for that, Jack, that would be great :ok_hand: 
screenshot from 2017-06-21 16-22-03

- [x] The arrows in the Geometry lesson 1 (5-7) need to be checked carefully for (the comments in the doc were a little confusing, so not sure if what I did was correct or not)

- [x] Arrows and Left and Right Cards resources have descriptions on both geometry lessons that need replacing (otherwise amusing place holder text will go in!).

After merging the following issues will need to be created:
- Use blank cards for job badges (rather than using an image that includes the text)
- Update thumbnail for job badges resource
- Use blank cards for left right cards (rather than using an image that includes the text)
- Update thumbnail for left right cards resource
- Update thumbnail for arrows resource
- Correct/check curriculum areas for the following learning outcomes:
    - unplugged-programming-give-instructions-shape
    - unplugged-programming-identify-bug-shape
    - unplugged-programming-describe-draw-2D-shape


behshidghorbani:
Can' remember if the "blocks needed" section should have the exact number of blocks or not repeat the ones that are the same and used more than once? These challenges all have one of each.

-All the Kidbots programs missing "learning outcomes".

-find-prime-factors: Testing examples outputs need ',' at the end.
Is it a good idea in the "Hints" section add a clickable link to those challenges?

-add-in-parts-2-digit-number: last paragraph in "Hints" change to:
To calculate the value of the variable “add to tidy num1”, subtract the last digit of “num1” from 10 (i.e. 10 - “num1” modulo 10). For example to make 74 a tidy number, you need to add 6 (10- (74 modulo 10)) making it a tidy number 80.

-catch-the-mouse-random/scratch-solution
remove "This is just one of many solutions" from the "view solution section.

-draw-regular-polygons
replace "There are no testing examples for this challenge" with "Test your program with 3, 5 and 10 number of sides as the input.".

-program-bee-bot
TRY IT OUT.
remove "This is just one of many solutions" from the "view solution section.

-add-in-parts-quiz
TRY IT OUT.

-draw-different-types-of-angles
Fix code "Hints" (red blocks) as discussed with Hayden.

-design-bee-bot
In the requirement section change: "A project with the graphics needed is provided, but you will need to write the program to make it work." to "A project with the graphics needed is provided at https://scratch.mit.edu/projects/159722851/, but you will need to write the program to make it work."
change the red blocks in "Hints" to green. (as discussed with Hayden)
Put each statement (forward, backward and so on) in a new line like this:
Forward: Adds the statement “forward” to the “source code” list. 
Backward: Adds the statement “backward” to the “source code” list. 
Right: Adds the statement “turn right” to the “source code” list. 
Left: Adds the statement “turn left” to the “source code” list. 
Pause: Adds the statement “pause” to the “source code” list. 
Undo: Removes the last item from the “source code” list. 
Clear: Removes all the items in the “source code” list. 
Home: Moves the “cat” sprite to position x:50, y:0 on the stage. Go: Executes all the statements in the “source code” list following below algorithm:

-calculate-inside-angles-of-regular-polygons
I don't think this is linked to the correct program. When click on the green flag it's not displaying the inside angle is ... output at the end.
replace "There are no testing examples for this challenge" with "Test your program with 3, 5 and 10 number of sides as the input.".

-draw-different-types-of-triangles
in the "Hints" section, add a link to challenge 6.1 (Draw different types of angles);
"In this challenge, the third side of the triangle can be drawn by connecting the third vertex (location of the sprite after drawing the angle in challenge 6.1) to the first vertex (x:0 y:0).

-9.1 reflection
has the wrong title (shouldn't be Draw 3D shapes (cube)).

-generate-primes
add ',' to the end of the outputs in the "testing examples".

-translation
has the wrong title (shouldn't be Draw 3D shapes (cube)).

-draw-star
The blocks needed are not correct. And should be 3 different sets for 3 different solutions?
In "Hints" can we add a link to Geometry challenge 7.2 in "Solution 1: By drawing a hexagon inside a star (see “Geometry challenge 7.2”)."?

-Find factors of a number
blocks needed solutions should be in seperate sections. They are both shown together so it's confusing.

PR #556, opened 2017-07-06
Searching algorithms

ravenmaster001:
Addresses issues:
- #547 
- #548 
- Lesson: _How many guesses?_
- Lesson: _Divide and conquer_
- Lesson: _The Great Treasure Hunt (Unsorted)_
- Lesson: _The Great Treasure Hunt (Sorted)_

Things to be fixed or changed **(input required)**:
- [x] Searching algorithms markdown file has no content.
- [x] Classroom resources text for lessons has been shortened, is this ok?
- [x] Missing curriculum areas for learning outcomes.

Issues to be raised from this **(input required)**:
- [x] Missing videos and text in Unit Plan (https://docs.google.com/document/d/14Exafr-nE7s4yDuvPMn-zv81gVvK5HMS8Unu-wRW2BI/edit) #561
- [x] All lessons: urls that point to www.example.com need to be fix, once generated resources are added.
- [x] Lesson 1: generated resources (https://docs.google.com/document/d/1Agn98IBahO1bEfnHaH_lPYYsjFxhws-bmK6KljMFI38/edit)
- [x] Lesson 2: generated resources (https://docs.google.com/document/d/1H1HUy887UPF2Ck75LGfE16SfQJtJTrMIc1zh2gZIlQ8/edit)
- [x] Lesson 3: generated resources (https://docs.google.com/document/d/1uN6RKx-Bn_G6lUnqA8cvw-W34jG0ocSHGvtDqX6AYOI/edit)
- [x] Lesson 4: generated resources (https://docs.google.com/document/d/16IndkdA5spbsFbqBwZBydRGXL0x4KEHZQ1454uqThfo/edit)
- [x] Treasure Hunt instructions pdf + links

PR #566, opened 2017-07-31
Add searching cards resource and tidy resource specfication

JackMorganNZ:
- Adds searching card resource.
- Adds HTML source of resource pages.
- Adds basic resource specification.
- Changes all resources to match new specification.

For merging into #556.

Issue should be raised for creating PDF CSS that doesn't trigger warnings.


Dockerfile-local
```diff
@@ -10,19 +10,37 @@ ARG DEBIAN_FRONTEND=noninteractive
...
       python3-pip \
-      libffi-dev \
-      libcairo2-dev \
       python-dev \
       python-pip \
       python-lxml \
       python-cffi \
-      libpango1.0-0 \
-      libgdk-pixbuf2.0-0 \
+      fontconfig \
```
ravenmaster001: Whoops, remove this.

JackMorganNZ: Which part are you referring to? I added the `fontconfig` for the fonts, but I can't remember anymore if it's still required (pretty sure it is).

Issue #572, opened 2017-08-01
Travis build exceeds maximum time


JackMorganNZ:
- https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts
- https://docs.travis-ci.com/user/speeding-up-the-build
- https://docs.travis-ci.com/user/build-stages/
- https://docs.travis-ci.com/user/build-stages/share-docker-image/

JackMorganNZ:
Progress started on branch `issue/572`.
Travis build showing build stages: https://travis-ci.org/uccser/cs-unplugged/builds/259968580

ravenmaster001:
Still failing: https://travis-ci.org/uccser/cs-unplugged/builds/261248392
Note the treasure hunt builds appear to be taking a ridiculous amount of time.

ravenmaster001:
@JackMorganNZ the deploy takes so long due to the amount of resources it is trying to build. My computer takes about 8mins for all the ones in the tests (single copy) and during deploy we ask for 30 copies. I'm thinking that resource deployment should be moved to a manual step for now.

JackMorganNZ:
If we make it a Travis build stage, does the job time reset?

ravenmaster001:
@JackMorganNZ Considering on my machine the command `time .\csu dev makeresources` reports back:
```
real	63m43.671s
user	0m0.196s
sys	0m0.044s
```
I'm not sure that it's going to matter.

JackMorganNZ:
Well that is an issue... Having 30 copies per resources ideally the minimum as that is a standard class sizes (though when I first asked educators, their answer was 100 copies!). Once we have the render system up and running this step would be skipped.

I might try one last solution instead of switching to manual deployment, but if it fails, manual will have to do!

JackMorganNZ:
Here are the times for generating resources on my machine:

```
Arrows                 : 2 seconds
Barcode Checksum Poster: 2 seconds
Binary Cards           : 23 seconds
Binary Cards (Small)   : 15 seconds
Binary to Alphabet     : 1 second
Binary Windows         : 18 seconds
Grid                   : 1 second
Job Badges             : 1 second
Left and Right Cards   : 1 seconds
Modulo Clock           : 2 seconds
Parity Cards           : 3 seconds
Piano Keys             : 3 seconds
Searching Cards        : 34 seconds
Sorting Network        : 47 seconds
Sorting Network Cards  : 14 seconds
Train Stations         : 1 second
[...]

Issue #425, opened 2017-06-01
Travis test backwards passes when it should not 


ravenmaster001:
Noticed in my builds that somehow `test_backwards` passes even though it probably shouldn't (as django cannot be run) see the Travis build here: https://travis-ci.org/uccser/cs-unplugged/builds/238211827

Issue #231, opened 2017-04-27
Automatically publish system Docker images to Docker Hub for easier installation


JackMorganNZ:
The required Django and Nginx images should be available on Docker Hub for easy installation by developers and content authors.

Travis CI should always build from scratch for error checking, and on a successful build on the `develop` branch (eventually `master`), it would deploy the images to Docker Hub.

The `csu` script would also need to be altered to search for these images, with an override for building using the local Dockerfiles.

Issue #203, opened 2017-04-24
Lock Python version in Docker image to specific version


JackMorganNZ:
Created from work in #193.

JackMorganNZ:
It might be worth switching to a base Python Docker image: https://hub.docker.com/_/python/

JackMorganNZ:
@ravenmaster This issue and comment above is related to what we were talking about last week.

PR #573, opened 2017-08-02
Travis Build Stages + Docker Repository

ravenmaster001:
Locks the python version #203 
Uses base images to reduce build time #231 
Fixes csu script to ensure correct status code for backward testing #425 
Attempts to address maximum timeouts #572


ravenmaster001:
Currently using travis stages for parallalisation of tasks, but it should be noted that having to start the system each time adds roughly ~12 minutes to the total time taken.

ravenmaster001:
Possibly chopped off 4~8 minutes looking at previous build (although the trend upwards before failed builds was quite rapid).

ravenmaster001:
@JackMorganNZ I think what I wanted to do here is done. Not sure this fixes the problem might be worth turning off resource generation tests off unless it is a pull request (example line is in csu waiting to be uncommented).

I found that resource generation tests take about ~8 minutes on my machine and considering they are essentially run again at deployment to make all the variants that is a significant portion of time. 

I haven't attempted anything here to speed up deployment scripts.

JackMorganNZ:
Would you be able to explain this more, don't quite get it sorry:

> Possibly chopped off 4~8 minutes looking at previous build (although the trend upwards before failed builds was quite rapid).

ravenmaster001:
@JackMorganNZ if you look at the travis build history a while ago the builds were about 10 minutes, then recently they quickly sped up to about 20+ minutes just before develop broke. I think the new resources are causing a (mild) kind of state explosion increasing the build times drastically.

PR #575, opened 2017-08-08
Made csu script more script-y

javanut13:
Made the csu script more extensible.

You can define new subcommands like:

```shell
cmd_foo() {
  echo 'Foo command'
}
defhelp foo 'This is the help for foo'
```

And a dev command like:

```shell
dev_bar() {
  echo 'Bar command'
}
defhelp -dev bar 'Help for bar'
```


csu
```diff
@@ -13,411 +13,294 @@
...
-  if [ $# -lt 1 ]
-  then
-    printf "${RED}ERROR: dev command requires one parameter!${NC}\n"
-    helpmenu
-    exit 1
+cmd_dev() {
+  local cmd="$1"
+  shift
+  if [ -z "$cmd" ]; then
+    echo -e "${RED}ERROR: dev command requires one parameter!${NC}"
```
ravenmaster001: Just need to check but does the `-e` here enable the `${RED}` modifier?

csu
```diff
@@ -13,411 +13,294 @@
...
+defhelp -dev test_coverage 'Display code coverage report.'
 
 # Run test suite backwards for CI testing
-function test_suite_backwards {
-  printf "Running test suite backwards...\n"
-  if [ "$TRAVIS_PULL_REQUEST" != "false" ]
-  then
+dev_test_suite_backwards() {
+  echo "Running test suite backwards..."
+  if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
ravenmaster001: You might encounter a merge conflict here. I added some stuff around this recently to make sure that when the command is not run under a pull request the result is 0, while if the command is run the result is that of the docker-compose command. The code might be superfluous, so please take time to consider it when merging it in.

csu
ravenmaster001: Answering my own question here, yes `-e` is necessary to match the `\033` escape character for the colour change.

csu
```diff
@@ -13,411 +13,294 @@
...
+  cmd_end
+  cmd_start
 }
+defhelp restart 'Stop and then restart development environment.'
 
 # Run Django migrate and updatedata commands
-function update {
-  static
+cmd_update() {
+  cmd_static
```
ravenmaster001: This should be `dev_static` not `cmd_static`

csu
```diff
@@ -13,411 +13,294 @@
...
 
-  printf "\n"
-  migrate
+  echo ""
+  cmd_migrate
```
ravenmaster001: Again `dev_migrate` not `cmd_migrate`

csu
```diff
@@ -13,411 +13,294 @@
...
 
-  printf "\n"
-  updatedata
+  echo ""
+  cmd_updatedata
```
ravenmaster001: `dev_updatedata` not `cmdupdatedata`

ravenmaster001:
Approving this, but changing the branch so I can do some minor changes based on the `ci` commands.

ravenmaster001:
Approving this, but changing the branch so I can do some minor changes based on the `ci` commands.

PR #627, opened 2017-10-17
Add canterbury.ac.nz domains to allowed hosts in local settings

jordangriffiths01:
This will allow developers working within the UC network to access django instances running on other developer's machines, via the <hostname>.canterbury.ac.nz.


JackMorganNZ:
A test failed, but was due to Travis network error. Have restarted it.

Will review tomorrow when on site.

PR #633, opened 2017-10-20
Implement extraction of translatable strings from topics application models

jordangriffiths01:


csunplugged/topics/management/commands/_LessonsLoader.py
```diff
@@ -139,7 +117,7 @@ def load(self):
                 slug=lesson_slug,
                 duration=lesson_duration,
                 languages=list(content_translations.keys()),
-                classroom_resources=classroom_resources,
+                # classroom_resources=classroom_resources,
```
JackMorganNZ: Delete commented line

jordangriffiths01:
There will need to be fairly significant changes to much of the documentation for many aspects of this internationalisation process - for example, the structure files are now stored under /structure not /en.

I think it will be easier to do this all together at a later stage, in a separate PR. Thoughts?

csunplugged/topics/models.py
```diff
@@ -427,3 +442,15 @@ class ResourceDescription(models.Model):
...
+
+
+# Register translatable strings populated from yaml files using vinaigrette
+vinaigrette.register(CurriculumArea, ["name"])
+vinaigrette.register(ProgrammingChallengeDifficulty, ["name"])
+vinaigrette.register(ProgrammingChallengeLanguage, ["name"])
+vinaigrette.register(AgeGroup, ['description'])
+vinaigrette.register(ResourceDescription, ['description'])
+vinaigrette.register(ClassroomResource, ['description'])
+# TODO: Implement translation of classroom resources
```
JackMorganNZ: Looks like these lines can be removed.

PR #645, opened 2017-10-26
Fix resource cache function (fixes #644)

### LINKED ISSUE #644: Resource cache returns 500 error
Bug only on `develop`.

JackMorganNZ:
Fixes a bug in the function for creating redirect URL to the resource cache. Also adds a test to check this function works correctly.


csunplugged/tests/resources/other/test_resource_cache_redirect.py
```diff
@@ -0,0 +1,30 @@
...
+        resource = self.test_data.create_resource(
+            "grid",
+            "Grid",
+            "resources/grid.html",
+            "GridResourceGenerator",
+        )
+        query = QueryDict("paper_size=a4")
+        generator = get_resource_generator(resource.generator_module, query)
+        response = resource_pdf_cache(resource.name, generator)
+        self.assertEqual(response.status_code, 302)
```
jordangriffiths01: Thoughts on having these status codes by name?

```
from http import HTTPStatus
self.assertEqual(response.status_code, HTTPStatus.FOUND)
```

PR #652, opened 2017-10-28
Modify csu script and docker setup for OSX compatability (fixes #651)

### LINKED ISSUE #651: Modify docker configuration to work on OSX
Docker for Mac does not properly support the `network_mode: host` option for containers. In order to run the system on OSX, it will be necessary to network the containers using a bridged network:

> By default Compose sets up a single network for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a hostname identical to the container name."

Rather than accessing other containers via a port on localhost, containers will  access each other using the instance name as the hostname. Port 80 will then be exposed from the nginx container to the host.

jordangriffiths01:
Modify docker containters to use the default bridged network, because
Docker for Mac does not support `network_mode: host`

Modify csu script to remove the use of `xargs -r` which is only
available on GNU (not available on OSX)

@JackMorganNZ, @hayleyavw: I've run this successfully on both my macbook and my ubuntu VM, but can you think of any potential issues this change could cause in production/elsewhere?


JackMorganNZ:
Sadly `network_mode: host` is required for two reasons:

1. **Deployment:** Internet access is required by the Docker images for deploying.
2. **Render system (coming soon):** @ravenmaster001 mentioned this will be required as the system will need to talk to the render system, however I am unsure of the actual implementation of this. I'm still fuzzy on if render can be implemented as a image/s within the local compose setup, or if it exists separately, especially as they are separate projects/repos so need to be in sync (for example, `develop` requests `develop`, etc).

The `docker-compose.yml` file is only used for local development, so we could create an alternative file for OSX. The following links may be useful if this is the best option:

- Set docker compose file: https://docs.docker.com/compose/reference/overview/
- Use multiple files or override file: https://docs.docker.com/compose/extends/

JackMorganNZ:
Closing while changes are made to deployment.

JackMorganNZ:
@jordangriffiths01 I've updated this to now use the Cloud SQL Proxy via Docker Compose, and checked it deployed correctly. Give it a check, and it's all good to go.

jordangriffiths01:
@JackMorganNZ changes you made look great, and I've checked the local dev environment still builds and functions on my MacBook Pro (it does). Thanks for your work on this!

PR #655, opened 2017-10-29
Collect resource thumbnails for static file deployment 

### LINKED ISSUE #653: Resource thumbnails are not deployed

JackMorganNZ:
Resource thumbnail images were not collected for deployment, therefore the scripts have been updated to  include the resource thumbnails. I also added clearer steps by deleting existing folders and creating production static files from a fresh slate, as Scratch image files were being deployed but accidentally as a remnant from the start up script.

Fixes #653

### DIFF
--- a/infrastructure/dev-deploy/deploy-static-files.sh
+++ b/infrastructure/dev-deploy/deploy-static-files.sh
@@ -5,7 +5,11 @@
 ./csu start
 
 # Generate production static files
+rm -r build/
+rm -r staticfiles/
 ./csu dev static_prod
+./csu dev static_scratch
+./csu dev makeresourcethumbnails
 ./csu dev collect_static
 
 # Install Google Cloud SDK
--- a/infrastructure/prod-deploy/deploy-static-files.sh
+++ b/infrastructure/prod-deploy/deploy-static-files.sh
@@ -5,7 +5,11 @@
 ./csu start
 
 # Generate production static files
+rm -r build/
+rm -r staticfiles/
 ./csu dev static_prod
+./csu dev static_scratch
+./csu dev makeresourcethumbnails
 ./csu dev collect_static
 
 # Install Google Cloud SDK


infrastructure/dev-deploy/deploy-static-files.sh
```diff
@@ -5,7 +5,11 @@
 ./csu start
 
 # Generate production static files
+rm -r build/
```
jordangriffiths01: This feels wrong, although I'm not super familiar with the script. This seems to do a build (not production), then delete the static files and regenerate them with production settings. That means running commands like `./csu makeresourcethumbnails` twice (once within csu update, called from csu start), and again here.

Is there a reason we can't replace 
```
./csu start
rm -r build
rm -r staticfiles
```
with 
```
docker-compose up  # could be wrapped in a csu function for maintainability
./csu dev migrate
./csu dev updatedata
```
or whatever commands are required?

It's very possible I'm missing something here, not being super familiar with the deployment process!

JackMorganNZ: That would make more sense, but sadly `updatedata` has a chicken and egg scenario.

- Need to run it after `static` so it has the static files available to detect that images are found, otherwise it throws an error.
- Need to run it before `static_scratch` as the Scratch text files need to exist for creating the images.

An option would be to run Verto with a flag to not raise errors if files aren't there, but this seems a bit dangerous (even though it would have passed many tests to get to the deployment stage).

JackMorganNZ: After checking the Verto and CS Unplugged repos (been ages since I've looked at this code), the errors raised for missing images happens in the CS Unplugged loader, so would be easier to override if this approach is taken.

JackMorganNZ: Have discussed with @jordangriffiths01, will use as short term solution and will revisit in new issue.

PR #658, opened 2017-10-30
I18n backend changes, and frontend prototype

jordangriffiths01:
This PR:
- Prototypes i18n changes to the front end.
- Extracts any strings out of structure yaml files into their own yaml files within each locale directory.
- Removes the use of vinaigrette to extract translatable strings from the database - instead, all translatable strings are now stored in their own yaml files in the respective locale directory tree, and loaded into their own database columns using the `django-modeltranslation` package. This consolidates our approach to translation of a model, and gives us much more control (`vinaigrette`, which uses `ugettext`, gave us no easy way to detect when translations were available).
- Introduces a new loader class, `TranslatableModelLoader`, which contains all functionality to load translated yaml/markdown files, and populate the relevant field on translatable models.
- Modifies existing loaders to subclass `TranslatableModelLoader`, to cut down on repeated code.
- Removes `null=true` option for all model `CharField`/`TextField`s, as recommended by django documentation. This is also required in order to prevent fallback to english on certain translated fields.
- Prevents fallback to English on certain fields using the `translation_undefined` attribute for `TranslationOptions` classes.
- Adds custom managers to the `TranslatableModel` class which allow for queries to retrieve only translated or untranslated instances of any TranslatableModel class.

It does not do the following (which will come in a later PR):
- Implement new tests for loaders as a result of removing `vinaigrette` and populating the translated strings in the loaders.
- Implement production ready front end changes - @JackMorganNZ is likely to build upon this prototype to implement the front end changes that will make it to production.

It also includes the following which **do not need** to be reviewed, and are included temporarily in the feature branch:
- translated markdown/yaml files (anything inside `de` directory trees) and translated `.po` files.
- temporary patch to allow development on OSX, which will be reverted before a PR to `develop`.


csunplugged/topics/models.py
```diff
@@ -71,12 +54,12 @@ class Meta:
...
 
 
-class LearningOutcome(models.Model):
+class LearningOutcome(TranslatableModel):
     """Model for learning outcome in database."""
 
     #  Auto-incrementing 'id' field is automatically set by Django
     slug = models.SlugField(max_length=80, unique=True)
-    text = models.CharField(max_length=200, unique=True)
+    text = models.CharField(max_length=200, default='')
```
JackMorganNZ: Is `default=''` required? Is this the same as `blank=True` (which I thought was default for string fields).

csunplugged/utils/TranslatableModelLoader.py
```diff
@@ -0,0 +1,163 @@
...
+                    <model-slug>: {
+                        <language>: {
+                            <field-name>:<value>
+                        }
+                    }
+                }
+
+        Raises:
+            CouldNotFindConfigFileError: the requested file could not be found in
+                the /en directory, raised only if required_slugs is not empty.
```
JackMorganNZ: Should this be `could not be found in the /structure directory`? I am slightly confused.

csunplugged/topics/models.py
JackMorganNZ: From our discussion on Slack, this is required for the translation package.

jordangriffiths01: TL;DR Yes, `default=''` is required.

I thought exactly the same when I first implemented it, and was baffled by the fact that, without `default=''`, I was getting back `None` on unpopulated fields rather than `''`. This was at odds with the documentation, which says that generally no value for a  `TextField`/`CharField` is represented by an empty string on the database level, instead of null.

Turns out, `django-modeltranslation` sets `null=True` on the extra fields it creates for each language (i.e. `content_en`, `content_de` will have null=True even if the original `content` did not). This means `default=''` is required to force those translation fields to store no value as the empty string rather than null.

csunplugged/utils/TranslatableModelLoader.py
jordangriffiths01: The `/en` is correct - the error is related to missing yaml files storing the english strings. But you're right to be confused - the exception name is misleading as it now applies to all yaml files, not just config files. I've updated the exception class to reflect this.

csunplugged/topics/models.py
jordangriffiths01: Relevant issue on django-modeltranslation discussing this behavior - deschler/django-modeltranslation#144

PR #667, opened 2017-11-06
Add pixel painter resource

JackMorganNZ:
This pull request has had major scope creep and includes the following changes:

- Pixel painter resource.
- Ability for resources to provide custom overrides for thumbnails.
- PDFs are now created within the `BaseResourceGenerator`.
- Thumbnails are now created and written within the `BaseResourceGenerator`.
- Moved location of `BaseResourceGenerator`.
- Move resources generators to `resources/generators/` as the modules don't provide views.
  - Current tests of resources are not moved from `views` directory as they test via different view calls. There is an example of a generator test for the pixel painter resource.
- Resources view file is now at top level of application as it's the only view file.
- Update documentation.


csunplugged/resources/utils/resize_encode_resource_images.py
```diff
@@ -0,0 +1,41 @@
+from PIL import Image
+from io import BytesIO
+import base64
+
+MM_TO_PIXEL_RATIO = 6
+
+
+def resize_encode_resource_images(generator, data):
```
jordangriffiths01: Probably better to only pass through the paper size here, not the whole generator object, to reduce coupling.

csunplugged/resources/management/commands/makeresourcethumbnails.py
```diff
@@ -47,36 +48,27 @@ def handle(self, *args, **options):
...
-    return html.write_png(stylesheets=[base_css], resolution=72)
+                thumbnail_file_path = os.path.join(base_path, filename)
+                thumbnail = generator.thumbnail()
+
+                if isinstance(thumbnail, list):
+                    from weasyprint import HTML, CSS
+                    context = dict()
+                    context["resource"] = resource.name
+                    context["paper_size"] = generator.requested_options["paper_size"]
+                    context["all_data"] = []
```
jordangriffiths01: Empty list here? This looks like it won't generate anything?

csunplugged/utils/BaseResourceGenerator.py
```diff
@@ -76,3 +79,29 @@ def process_requested_options(self, requested_options):
                 values[i] = update_value
             requested_options.setlist(option, values)
         return requested_options
+
+    def thumbnail(self):
```
jordangriffiths01: I'd suggest you make this a function `generate_thumbnail(self, path)` which will actually write out the image file. The 'override' mechanism can then become the subclass overloading this function, to save a pregenerated PNG rather than dealing with pdf generation and conversion to png. The subclass implementation can choose to ignore the requested options, and generate the same png for every combination, if it so chooses.

This would mean the logic in lines 59-72 in makeresourcethumbnails.py will move into this function.

JackMorganNZ: Would this mean the same should happen for making a resource PDF?

csunplugged/resources/utils/BaseResourceGenerator.py
```diff
@@ -0,0 +1,187 @@
...
+            resource_name: Name of the resource (str).
+            path: The path to write the thumbnail to (str).
+        """
+        # Only import weasyprint when required as production environment
+        # does not have it installed.
+        from weasyprint import HTML, CSS
+        context = dict()
+        context["resource"] = resource_name
+        context["paper_size"] = self.requested_options["paper_size"]
+        context["all_data"] = [[self.generate_thumbnail]]
```
jordangriffiths01: `thumbnail_data` doesn't get referenced at all in this function - instead it looks like you're calling out to `self.generate_thumbnail` here. Also seems strange that you're passing the function through rather than calling it - is that intentional?

csunplugged/resources/utils/resize_encode_resource_images.py
```diff
@@ -0,0 +1,42 @@
...
+    Args:
+        paper_size: Paper size of requested resource (str).
+        data: List of generated resource (list).
+
+    Returns:
+        List of processed resource pages.
+    """
+    if paper_size == "a4":
+        max_pixel_height = 267 * MM_TO_PIXEL_RATIO
+    else:  # Letter paper size
```
jordangriffiths01: This seems like a big assumption to make. Consider
```
elif paper_size == "letter":
    # Stuff
else:
    # raise exception
```

PR #677, opened 2017-11-09
Improve project test coverage

JackMorganNZ:
This pull request adds much needed tests to areas of the system which had zero tests. This pull request also:

- Removes unused modules/logic.
- Moves several paths that were defined in modules to the settings file.
- Creates resource view tester.
- Moves testing only modules from `utils` to `tests`.
- Fixes various bugs.

There is still many areas of the system which are poorly tested, in particular the topic loaders, which still have areas where input is not properly tested.

The remaining areas of the system that have not had test coverage added/improved are:
- `TextBoxDrawer.py` - More efficient for author of module to design tests.
- Production and alternative settings and `wsgi` module - Need to investigate how to test these settings (will raise issues for this).
- ~`loadtopics` - This is being heavily rewritten in the upcoming i18n branch, so will develop tests after.~ Tests now exist for this command.

Fixes #32.
Fixes #165.
Fixes #167.
Fixes #346.

Issue #622, opened 2017-10-13
Implement automatic upload to Crowdin of .po file(s) on develop branch


jordangriffiths01:
Any changes to any .po files (probably updated by Travis when any static translatable strings change, see #621) should be synchronised to the Crowdin server.

Any new translations completed on Crowdin to those files should trigger a pull request back into `develop` with the translated versions.

Issue #621, opened 2017-10-13
Implement automatic update of .po file 


jordangriffiths01:
The .po file needs to be updated with all static strings for translation, by running the `python manage.py makemessages`. This needs to be run upon any change to templates or database content.

This process should be automated on Travis to run when any such files are updated on `develop`.

Issue #620, opened 2017-10-13
Implement automatic upload of new content file on Develop to Crowdin


jordangriffiths01:
When new/updated content markdown files are merged onto `master`, these need to be uploaded to Crowdin. 

While Crowdin ships with GitHub integration, this is not suitable for this use case (see #619). Instead, this will probably need to be implemented as a Travis job that runs for new commits to `master`. Upload can be achieved through the Crowdin CLI (https://support.crowdin.com/cli-tool/).

jordangriffiths01:
After much discussion, it has been decided to upload new content from `develop` rather than `master`. This is because `develop` is essentially treated as a pre-release staging branch, and we will be merging new translations into `develop` so keeping it in one branch is cleaner.

PR #680, opened 2017-11-11
Translation Infrastructure

### LINKED ISSUE #611: Update all templates to include translation tags around translatable strings
All templates must be updated to include {% trans ""%} or {% blocktrans %} tags around any static strings that require translation. 

This allows them to be extracted into a .po file for translation using the makemessages command, and then dynamically translated into a user's language.

jordangriffiths01:
This pull request adds translation support to the backend of csunplugged. This required some major changes, particularly to the models and loaders. Many of the changes have already had an initial pass by @JackMorganNZ (on PRs into the `translation` branch), but this PR will need a very thorough review due to the extensive and wide-reaching nature of these changes.

This PR makes no user-facing changes - an end user should not notice any change when interacting with the site. During development for this PR, prototype front end changes were made to support i18n, and this is intended to be used as a starting point for future development of production-ready front end modifications. This prototype has been reverted for this PR, but is available by checking out commit 078af51d8c9c9d21a26d2fbce5d66eaee1ed0e66.

Overview of major changes:
- Use of django-modeltranslation package to support translations of model fields by creating new database columns
- Add new abstract base model class `TranslatableModel` which is subclassed by any model with translatable content
- Moving structure `.yaml` files out of the `en` directory tree into a new directory, `structure`
- Extract translatable strings from structure `.yaml` files into their own `yaml` files, which are located in each languages directory tree
- Refactoring of loaders to handle the new directory structure and reduce code duplication
- Add new loader base class - `TranslatableModelLoader` - which handles population of translated content, and determining the availability of a model in a certain language based on the presence/absence of translations of particular fields
- Modify existing loaders to subclass `TranslatableModelLoader` and populate translated content
- Add new `ClassroomResource` model to replace raw text strings in `.yaml` file
- Unit testing of added i18n functionality, along with more comprehensive testing of existing loader functionality (@JackMorganNZ this will have some conflicts with #677, but with minor tweaks the changes in that PR will further improve testing of this area of the codebase)
- Add django `{% trans %}`/`{% blocktrans %}` tags to all translatable strings on the front end
- Add basic support for content upload to Crowdin translation service - this will be extended in later PRs

Aside from these major changes, there were many, many smaller changes to various parts of the back end as determined necessary.

This PR does not:
- Include any user-visible front end changes - development of a production-ready front end will come in a later PR (tracked by #616, #617)
- Add any translated content
- Implement automation of translation push/pull with Crowdin (tracked by #619, #620, #622)
- Implement automatic generation of .po translation files (tracked by #621)
- Add translation support for the resources application - the core issues with resource translation were solved with TextBoxDrawer (#670), but there are still some issues to work through with regards to resource versioning, and so this will follow in a later PR
- Update documentation - this will come in a later PR (tracked by #681 and #682)

Fixes #611, #612, #613, #614, #615


csunplugged/tests/resources/loaders/test_resource_loader.py
```diff
@@ -2,9 +2,9 @@
 from tests.resources.ResourcesTestDataGenerator import ResourcesTestDataGenerator
 from resources.models import Resource
 from resources.management.commands._ResourcesLoader import ResourcesLoader
-from utils.errors.CouldNotFindConfigFileError import CouldNotFindConfigFileError
+from utils.errors.CouldNotFindYAMLFileError import CouldNotFindYAMLFileError
```
JackMorganNZ: This was 'Config' instead of YAML to make it easier for authors to identify, however the file extension is `.yaml` and the filepath is displayed in the message, so I'm happy for it to be changed.

csunplugged/tests/utils/translatable_model_loader/test_TranslatableModelLoader.py
```diff
@@ -0,0 +1,260 @@
...
+    """Test class for TranslatableModelLoader."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.base_path = "tests/utils/translatable_model_loader/assets"
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        imp.reload(mt_settings)  # Let modeltranslation pick up any overriden settings
```
JackMorganNZ: "overriden" to "overridden"

csunplugged/topics/management/commands/loadtopics.py
```diff
@@ -21,12 +21,14 @@ def handle(self, *args, **options):
                 attribute.
         """
         factory = LoaderFactory()
+
         # Get structure and content files
         base_loader = BaseLoader()
-        BASE_PATH = "topics/content/en/"
+        base_path = "topics/content/"
```
JackMorganNZ: Could this be moved to the settings file? I presume this would then make it easier to test `loadtopics` which isn't currently tested.

csunplugged/topics/views.py
```diff
@@ -345,6 +350,17 @@ def get_queryset(self):
         """
         return GlossaryTerm.objects.order_by("term")
 
+    def get_context_data(self):
+        """Get context data for template rendering."""
+        return {
+            "glossary_terms": GlossaryTerm.objects.filter(
+                Q(languages__contains=[get_language()])
+            ).order_by("term_en"),
+            "untranslated_glossary_terms": GlossaryTerm.objects.filter(
```
JackMorganNZ: Why is `untranslated_glossary_terms` added to the context?

csunplugged/utils/TranslatableModel.py
```diff
@@ -0,0 +1,46 @@
...
+        """Get query set for untranslated objects."""
+        return super().get_queryset().filter(
+            ~Q(languages__contains=[get_language()])
+        )
+
+
+class TranslatableModel(models.Model):
+    """Abstract base class for models needing to store list of available languages."""
+
+    languages = ArrayField(models.CharField(max_length=5), size=100, default=[])
```
JackMorganNZ: Why is the size stated? Django documentation states PostgreSQL doesn't enforce this yet, but this argument is optional.

crowdin_content.yaml
```diff
@@ -0,0 +1,32 @@
+#
+# Your crowdin's credentials
+#
+"project_identifier": "cs-unplugged"
```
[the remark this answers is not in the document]
jordangriffiths01: The skeleton of this file was automatically generated using `crowdin generate` (see https://support.crowdin.com/cli-tool/), which created some with quotes and some without. 

I only modified certain values as required, and figured I'd keep everything else as it was generated. Happy to change if you disagree?

crowdin_content.yaml
```diff
@@ -0,0 +1,32 @@
...
+#
+files: [
+    {
+        "source": "/csunplugged/topics/content/en/**/*.yaml",
+        "translation": "/csunplugged/topics/content/%two_letters_code%/**/%original_file_name%",
+
+        # The parameter "update_option" is optional. If it is not set, translations for changed strings will be lost. Useful for typo fixes and minor changes in source strings.
+        # e.g. "update_as_unapproved" or "update_without_changes"
+        #
+        #"update_option" : "",
```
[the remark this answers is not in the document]
jordangriffiths01: This was part of the generated skeleton - I decided to leave it here because this is an option that we are likely to use, and it's useful having a reminder that it is available. 

Happy to remove if you disagree.

crowdin_content.yaml
```diff
@@ -0,0 +1,32 @@
+#
+# Your crowdin's credentials
+#
+"project_identifier": "cs-unplugged"
+"api_key_env": CROWDIN_API_KEY
+"base_path": ""
+#"base_url" : ""
```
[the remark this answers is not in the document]
jordangriffiths01: Yeap, done.

csunplugged/resources/management/commands/_ResourcesLoader.py
```diff
@@ -14,15 +14,9 @@
...
-    def __init__(self, structure_file, BASE_PATH):
-        """Create the loader for loading resources.
-
-        Args:
-            structure_file: file path for structure YAML file (str).
-            BASE_PATH: base file path (str).
-        """
-        super().__init__(BASE_PATH)
-        self.structure_file_path = os.path.join(self.BASE_PATH, structure_file)
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
jordangriffiths01: Good spotting - must have done something at some point but no longer required

csunplugged/topics/management/commands/loadtopics.py
jordangriffiths01: Agreed, however I suggest this is done at the same time as writing tests for `loadtopics.py`, which should be in a separate PR after #677 is merged (or as part of that PR).

csunplugged/topics/views.py
jordangriffiths01: As part of the UI prototype, the glossary page showed translated terms, followed by untranslated terms in English, under a warning banner. 

Something similar will be required for the final UI, so removing it would only be temporary. Leaving it here doesn't change the user experience aside from a very minor increase in response time.

Thoughts?

csunplugged/utils/TranslatableModel.py
jordangriffiths01: Removed, I didn't realise this was optional.

csunplugged/config/settings/testing.py
```diff
@@ -71,3 +71,15 @@
...
     "dev.apps.DevConfig",
 ]
+
+# Save production LANGUAGES setting in case any tests need to refer to it
+PRODUCTION_LANGUAGES = LANGUAGES  # noqa: F405
+
+# Override production value of LANGUAGES - this is what django-modeltranslation
+# will use when adding translated fields to the models on startup, which is
+# necessary to test i18n backend functionality.
+LANGUAGES = (
```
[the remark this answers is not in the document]
jordangriffiths01: Yes, it really would be ideal to set this on a per-test basis - this is what I tried to achieve initially. Unfortunately, a change in the available languages requires a change to the database schema because modeltranslation has to add/remove database columns for each language for a translatable field.

Overriding on a per-test-class basis works for `test_TranslatableModelLoader.py` because it never touches the database, although some trickery was still needed in the `setUpClass` method to get modeltranslation to pick up the new settings (see 8ceadf61bda145a2afe4025a17e300626605c298). That being said, despite being possible, I'm not currently overriding the LANGUAGES setting in that test class.

All the other loader tests are actually loading models into the db, and so would require schema modifications for each test case. This actually is possible - the unit tests for the `django-modeltranslation` package do exactly this - but it's messy (see https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/tests/tests.py#L110) and I believe this approach to be overkill for our purposes.

Setting LANGUAGES in the testing.py settings file was the best/simplest compromise I could come up with, but I'm open to other suggestions.

crowdin_content.yaml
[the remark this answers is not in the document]
JackMorganNZ: Righto, I'm leaning to cleaning up the file to keep the consistent tidy quality of the project.

csunplugged/config/templatetags/translate_url.py
```diff
@@ -0,0 +1,26 @@
+"""Module for the custom translate_url template tag."""
```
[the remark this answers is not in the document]
jordangriffiths01: This obviously slipped through the cracks - added now

csunplugged/config/settings/testing.py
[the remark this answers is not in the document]
JackMorganNZ: After reading all that, I definitely agree!

csunplugged/topics/views.py
JackMorganNZ: Happy to leave it here.

hayleyavw:
are there plans to update the documentation?

jordangriffiths01:
@hayleyavw absolutely - documentation is a priority in the next week or so, in a separate PR (tracked by #681 and #682).

csunplugged/tests/topics/loaders/test_curriculum_integrations_loader.py
```diff
@@ -12,20 +14,45 @@ class CurriculumIntegrationsLoaderTest(BaseTestWithDB):
...
-        ci_loader = CurriculumIntegrationsLoader(config_file, topic, self.test_data.LOADER_ASSET_PATH)
+        ci_loader = CurriculumIntegrationsLoader(topic, base_path=self.base_path, structure_filename=config_file)
         ci_loader.load()
 
         ci_objects = CurriculumIntegration.objects.all()
 
-        self.assertQuerysetEqual(
-            ci_objects,
-            ["<CurriculumIntegration: Integration 1>"]
-        )
```
hayleyavw: why are we removing this check that Queryset is as expected?

csunplugged/tests/topics/loaders/test_glossary_terms_loader.py
```diff
@@ -13,19 +14,39 @@ class GlossaryTermsLoaderTest(BaseTestWithDB):
...
+        translated_term = GlossaryTerm.objects.get(slug="glossary-term-1")
+        untranslated_term = GlossaryTerm.objects.get(slug="glossary-term-2")
+
+        self.assertSetEqual(set(["en", "de"]), set(translated_term.languages))
+        self.assertSetEqual(set(["en"]), set(untranslated_term.languages))
 
-        self.assertQuerysetEqual(
-            glossary_objects,
-            ["<GlossaryTerm: Glossary Term 1>"]
-        )
```
hayleyavw: why are we removing this check that Queryset is as expected?

csunplugged/topics/management/commands/_CurriculumIntegrationsLoader.py
```diff
@@ -1,27 +1,25 @@
...
+class CurriculumIntegrationsLoader(TranslatableModelLoader):
     """Custom loader for loading curriculum integrations."""
 
-    def __init__(self, structure_file_path, topic, BASE_PATH):
+    def __init__(self, topic, **kwargs):
         """Create the loader for loading curriculum integrations.
 
         Args:
-            structure_file_path: File path for structure YAML file (str).
             topic: Object of related topic model (Topic).
```
hayleyavw: missing `**kwargs` in args list

csunplugged/topics/management/commands/_LessonsLoader.py
```diff
@@ -1,36 +1,33 @@
...
     """Custom loader for loading lessons."""
 
-    def __init__(self, lessons_structure_file_path, topic, unit_plan, BASE_PATH):
+    def __init__(self, topic, unit_plan, **kwargs):
         """Create the loader for loading lessons.
 
         Args:
-            lessons_structure_file_path: file path to lessons yaml file (str).
             topic: Object of Topic model (Topic).
             unit_plan: Object of UnitPlan model (UnitPlan).
```
hayleyavw: missing `**kwargs` in args list

csunplugged/topics/management/commands/_ProgrammingChallengesLoader.py
```diff
@@ -15,20 +15,16 @@
...
+class ProgrammingChallengesLoader(TranslatableModelLoader):
     """Custom loader for loading programming challenges."""
 
-    def __init__(self, structure_file_path, topic, BASE_PATH):
+    def __init__(self, topic, **kwargs):
         """Create the loader for loading programming challenges.
 
         Args:
-            structure_file_path: File path for structure YAML file (str).
             topic: Object of related topic model (Topic).
```
hayleyavw: missing `**kwargs` in args list

csunplugged/topics/management/commands/_ProgrammingChallengesStructureLoader.py
```diff
@@ -32,79 +18,83 @@ def load(self):
...
             )
 
-            new_language.save()
+            translations = prog_languages_translations.get(prog_language, dict())
+            self.populate_translations(new_prog_language, translations)
+            self.mark_translation_availability(new_prog_language, required_fields=["name"])
+            new_prog_language.save()
 
-            self.log("Added programming langauge: {}".format(new_language.__str__()))
+            self.log("Added programming langauge: {}".format(new_prog_language.__str__()))
```
hayleyavw: omg how have we not noticed this typo in the output before? 😆  `langauge` -> `language`

csunplugged/topics/management/commands/_TopicLoader.py
```diff
@@ -2,31 +2,25 @@
...
-class TopicLoader(BaseLoader):
+class TopicLoader(TranslatableModelLoader):
     """Custom loader for loading a topic."""
 
-    def __init__(self, factory, structure_file_path, BASE_PATH):
+    def __init__(self, factory, **kwargs):
         """Create the loader for loading a topic.
 
         Args:
             factory: LoaderFactory object for creating loaders (LoaderFactory).
```
hayleyavw: missing `**kwargs` in args list

csunplugged/topics/management/commands/_UnitPlanLoader.py
```diff
@@ -2,35 +2,32 @@
...
     """Custom loader for loading unit plans."""
 
-    def __init__(self, factory, structure_file_path, topic, BASE_PATH):
+    def __init__(self, factory, topic, **kwargs):
         """Create the loader for loading unit plans.
 
         Args:
             factory: LoaderFactory object for creating loaders (LoaderFactory).
-            structure_file_path: File path for structure YAML file (str).
             topic: Object of related topic model (Topic).
```
hayleyavw: missing `**kwargs` in args list

csunplugged/utils/TranslatableModelLoader.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def get_yaml_translations(self, filename, field_map=None, required_slugs=[], required_fields=[]):
+        """Get a dictionary of translations for the given filename.
+
+        Yaml files must be structured as follows:
+            <object-slug>:
+                <field>: <translated value for field 2>
+                <field2>: <translated value for field 2>
+            <object-slug-2>
+                ...
```
hayleyavw: I think this note about the required yaml structure should be removed, this is the kind of thing that should be in the docs

csunplugged/utils/TranslatableModelLoader.py
```diff
@@ -0,0 +1,166 @@
...
+                    },...
+                }
+        """
+        for language, values_dict in model_translations_dict.items():
+            with translation.override(language):
+                for field, value in values_dict.items():
+                    setattr(model, field, value)
+
+    @staticmethod
+    def mark_translation_availability(model, required_fields=[]):
```
hayleyavw: why does `required_fields` have a default value/is there a situation where this method can be called without any fields needing translation?

csunplugged/tests/topics/loaders/test_curriculum_integrations_loader.py
jordangriffiths01: The test has essentially just been broken into more detailed checks in the lines below - I check only one object has been returned, and that the heading and content are as expected.

csunplugged/tests/topics/loaders/test_glossary_terms_loader.py
jordangriffiths01: See above.

csunplugged/topics/management/commands/_CurriculumIntegrationsLoader.py
jordangriffiths01: This was discussed offline with the whole team, but including here for reference - we have decided to leave this as is, so that superclass kwargs are only documented on the superclass. This is simpler, avoids duplication and is more maintainable.

csunplugged/topics/management/commands/_ProgrammingChallengesStructureLoader.py
jordangriffiths01: Good spotting :laughing:

csunplugged/utils/TranslatableModelLoader.py
jordangriffiths01: The required yaml structure will definitely be included in the docs, but I think it's also important here. I see them as serving different purposes.

- The documentation will include it for reference to anyone wishing to create new yaml files (i.e. authors)
- The function docstring outlines the type of YAML file that *this* function handles. If the file structure changes, this function will need to be updated to support that new structure (at which point the docstring would be changed to represent the new functionality).

Thoughts?

csunplugged/utils/TranslatableModelLoader.py
jordangriffiths01: As discussed offline, an empty list here will mean the model is available in all languages, regardless of whether any of its fields are translated. While I can't currently think of a good reason to use this, preventing it seems needlessly restrictive.

I've updated to docstring to make this clearer.

csunplugged/utils/TranslatableModelLoader.py
JackMorganNZ: @jordangriffiths01 Thanks for points out the differences between the two, however we don't do this for all the standard loaders (for example: this is what a lesson YAML looks like, so should we just do it in the documentation for consistency. The docstring could reference the docs also. Thoughts?

JackMorganNZ: Just chatted to @jordangriffiths01 about this. I took a lot of convincing but I'm fine for this to stay in the docstring.

My initial complaint was this was inconsistent with the standard loaders, however this method deals with YAML files in a very generic sense (there can be any number of objects, with their specific fields (of varying numbers). Therefore the documentation can detail the specifics of how to create these files for authors.

For developers, this is a crucial, but still confusing method (I still get confused by it even after reading the docstring) so having the detailed docstring plus the test cases is probably the best way to provide a clear understand of the method. Better to have extra, than less.

But the main point is that these files will also be detailed in the docs, while this snippet is to aid developers.

PR #699, opened 2017-11-22
Issue/694

### LINKED ISSUE #694: Fix migration handling on travis
Unit tests should run without migrations, with the the test database schema being determined entirely from the current state of the model files and the test settings.

Since merging #643 this has happened locally, but the tests running on Travis still build the schema using the migrations files. This causes issues when the current production database schema differs from that required for the unit tests (specifically, this is occurring for unit tests that test internationalisation features, when the production database only contains columns for English values). This actually should have prevented merging of #680, except ironically that branch was missing a migration file that removed old `de` and `fr` columns (created during feature development) from the database, and so they _were_ available to the unit tests. I won't be able to add that migration until this issue is resolved, as all i18n unit tests will fail until they are allowed to run without migrations.

Production database migrations should _still_ be tested - this basically requires booting the system and then running `./csu update` to check that everything loads correctly with no errors, essentially a dry run of the database update step in the deployment phase. 

Currently this actually happens at the start of _every_ travis job as part of the `./csu start` command. This seems overkill to me, and adds a massive overhead to each job needing to generate all thumbnails etc before actually running any tests. Instead, I'd suggest we refactor the csu script to separate out the 2 distinct phases that currently are both run as part of `./csu start`:
- Starting the docker containers (IMO, should be the only function of `./csu start`)
- Loading database content and other tasks such as static file handling, thumbnail generation etc. (`./csu update`, run from `./csu start`)

I propose the following changes:
- Remove the `./csu update` call from `./csu start`.
- Run all unit tests without migrations (both on travis and locally)
- Add a new travis job to run `./csu update` after `./csu start`, acting as a dry run of the database update step from deployment
- Add missing topics migration file to remove erroneous `de`/`fr` columns

Thoughts @JackMorganNZ @hayleyavw?

jordangriffiths01:
Fixes #694


.travis.yml
```diff
@@ -15,6 +15,7 @@ install:
 jobs:
   include:
     - stage: test
+      script: ./csu ci load_content
       script: ./csu ci test_general
```
JackMorganNZ: Would this be better in a distinct test from `load_content`?

csu
```diff
@@ -357,8 +354,14 @@ ci_docs() {
   dev_docs
 }
 
+ci_load_content() {
+  cmd_update
+}
+
 cmd_ci() {
   cmd_start
+  dev_static
```
JackMorganNZ: Would it be better to do this within the tests that require it? For example, style checker doesn't require this.

csunplugged/tests/resources/management/test_makeresourcethumbnails_command.py
```diff
@@ -22,9 +22,10 @@ def test_makeresourcethumbnails_command_single_resource(self):
             "resources/grid.html",
             "GridResourceGenerator",
         )
+        # TODO: Fix these tests, they shouldn't be writing files into the build directory
```
JackMorganNZ: Make an issue for this and I can sort this out.

csunplugged/utils/TranslatableModel.py
```diff
@@ -29,7 +29,7 @@ def get_queryset(self):
 class TranslatableModel(models.Model):
     """Abstract base class for models needing to store list of available languages."""
 
-    languages = ArrayField(models.CharField(max_length=5), default=[])
+    languages = ArrayField(models.CharField(max_length=10), default=[])
```
JackMorganNZ: Reason for this increase?

.travis.yml
jordangriffiths01: `load_content` is the test that I've added, and I believe it is run as a distinct test in the same stage (`test`). Correct me if I'm wrong.

csunplugged/utils/TranslatableModel.py
jordangriffiths01: Not all language codes are xx or xx_yy format (eg. Chinese simplified: zh_Hans). None supported by django are longer than 10 (see https://github.com/django/django/tree/master/django/conf/locale).

Catching it before it becomes a problem later.

csunplugged/tests/resources/management/test_makeresourcethumbnails_command.py
jordangriffiths01: Tracked by #700

.travis.yml
JackMorganNZ: This will run `load_content` with `test_general`, just change line 19 to be:
```
    - script: ./csu ci test_general
```
Much like the lines following it. Each script in the same stage is a new item of a different key. Very un-intuitive YAML file!

jordangriffiths01: Ohhhh you're talking about the missing `-`, with you now. That was just a typo haha

PR #707, opened 2017-11-24
Update navbar and footer

saranglove:


csunplugged/static/scss/website.scss
```diff
@@ -191,22 +191,29 @@ $rounded-corner-radius: 0.5rem;
...
   #navbar-brand-logo {
-    height: 2.5rem;
+    height: 2rem;
   }
 
-  .nav-link:hover {
-    text-decoration: underline;
+  .navbar-nav .nav-link {
+    color: $white;
+    &:hover {
```
JackMorganNZ: Should also include `:focus` (which is currently set to red).

csunplugged/templates/base.html
```diff
@@ -38,8 +38,8 @@
           <span class="navbar-toggler-icon"></span>
         </button>
         <a class="navbar-brand" href="{% url 'general:home' %}">
-          <img src="{% static 'img/logo-small.png' %}" id="navbar-brand-logo" class="d-inline-block align-top" alt="CS Unplugged logo">
-          <a href="http://cs-unplugged.readthedocs.io/en/latest/changelog.html" class="badge badge-red">{{ VERSION_NUMBER_ENGLISH }}</a>
+          <img src="{% static 'img/logo-mono.png' %}" id="navbar-brand-logo" class="d-inline-block align-top" alt="CS Unplugged logo">
```
JackMorganNZ: Shall we try exporting the logo at a small resolution? My browser is currently displaying it as `330x120` so 400px wide should be good.

Could also be an SVG? Then the colour could be applied as a fill? (see Khan Academy).

PR #714, opened 2017-11-26
Fix loading of deployment config variables; Switch BaseLoader to use locale code

jordangriffiths01:
3 things fixed in this PR:
- Loading of environment variables for deployment configuration was broken - this is fixed (hopefully).
- Message files (.po) are automatically compiled to .mo on deployment
- `BaseLoader` is modified to use locale code rather than language code when searching for a translation directory. This is so that content directories are consistent with `/locale`.


csunplugged/docker-development-entrypoint.sh
```diff
@@ -19,6 +19,9 @@ done
 
 >&2 echo "Postgres is up - continuing..."
 
+echo "Compiling message files"
```
JackMorganNZ: Should this happen here or in `csu update`, what is better for developers?

jordangriffiths01: As discussed offline, decided to leave here, because compiling message files requires a django reboot

JackMorganNZ: Agreed over phone this is suitable.

PR #731, opened 2017-11-28
Crowdin bot improvements

jordangriffiths01:
This PR makes the following changes to `crowdin-bot`:

- Modify pull-translations script to commit all `.po` files together onto a separate branch, rather than on the per-language translation branch with other content. This is because these files are pulled as soon as the source file is updated, rather than once translation is completed. Committing all onto a separate branch together greatly reduces the developer time required to merge these changes in.

- Add automatic updating of the source `django.po` file when new translatable strings are added to templates/python code.

- Add `--no-location` flag to makemessages command, which prevents context comments in resulting .po file

- Prevent commit/push/PR for trivial changes to `.po` files that only involve created/modified timestamp changes


infrastructure/crowdin/setup-instance.sh
```diff
@@ -8,11 +8,14 @@ PYTHON_PACKAGE_DIR="crowdin_bot_python_package"
 
 # Install packages
 sudo apt-get install git --yes
+sudo apt-get install gettext --yes
 sudo apt-get install default-jre --yes  # Java, for crowdin cli
 sudo apt-get install python3-pip --yes
 sudo apt-get install python3-lxml --yes  # Install here instead of pip3 because compilation uses too much RAM
 sudo pip3 install -U pip setuptools
```
JackMorganNZ: Could these files be stored in a requirements file to lock to same versions as the rest of the system, which will be automatically updated by PyUP with the same system.

infrastructure/crowdin/setup-instance.sh
```diff
@@ -8,11 +8,14 @@ PYTHON_PACKAGE_DIR="crowdin_bot_python_package"
 
 # Install packages
 sudo apt-get install git --yes
+sudo apt-get install gettext --yes
 sudo apt-get install default-jre --yes  # Java, for crowdin cli
 sudo apt-get install python3-pip --yes
 sudo apt-get install python3-lxml --yes  # Install here instead of pip3 because compilation uses too much RAM
```
JackMorganNZ: Should these be version locked? Or are we fine as is, especially as this is not critical infrastructure.

jordangriffiths01: I've added a requirements file for the python dependencies. I assume those are the ones you meant, not the apt-get packages :)

PR #760, opened 2017-12-14
Redesign of topic and unit plan webpages

### LINKED ISSUE #696: Topics homepage interface redesign
Current cards:
image

Proposed redesign (wireframe):
image

This would eliminate the extra step of going into a topic's page in order to access UP/lessons.

Current flow:
image
(Where the page circled in red is unnecessary)

Proposed flow:
image
(Where the step circled in blue would replace the topic's homepage, and the unit plan would be separated to a different page.)

**Content structure may be needed for topics with more than one unit plan.

### LINKED ISSUE #697: Topics page (Lessons page) Redesign
Currently, the topics page looks like:
https://cs-unplugged-dev.appspot.com/en/topics/binary-numbers/

To fix:
1. Unit plan link can be removed as the unit plan page can be accessed via the topics card on the topics homepage ( after issue #696 ) The topic's page will then just be the lesosn's page.
Current:
image
Proposed:
image

2. Lessons table can be remade to show one table with multiple tabs depending on age group.
Proposed (wireframe):
image

JackMorganNZ:
## Changes

- Simplify topic page by hiding lessons of unit plan (to reinforce that lessons are within unit plans)
- Display unit plan description on sub-page from unit plan page to not hide lessons at end of page. A truncated version of description is provided on the unit plan page.
- Use two column design for topic and unit plan pages on larger screen devices.
- Soften hover state of `link-item` by changing background colour from red to blue.
- Remove underline from headings to clean up pages.
- Add helper function for calculating lesson range for topic and unit plan objects.
- Tidy start of unit plan documents (remove level 2 heading directly after level 1 heading).
- Fix bug where lesson duration wasn't shown.
- Fix bug where language dropdown was aligned in opposite direction.
- Greatly simplify `render_html_field` tag.

Fixes #696 
Fixes #697

## Reviewers

- @saranglove: For design changes
- @hayleyavw / @jordangriffiths01: For technical changes


hayleyavw:
It's looking so much better! 😍 
Technical side of things look LGTM, I did notice one design thing when taking a peak around the site,
when viewing a unit plan on a "large screen" I immediately noticed that the space between the main heading and the one below feels a bit close.
screen shot 2017-12-14 at 6 14 38 pm

Also not a graphic designer though....

JackMorganNZ:
Ready for review again @hayleyavw @saranglove. I added a bit of padding between the headers (we don't often have two headings against each other hence why it looked tight compared to other pages), and I updated the tests to all use kwargs (must have been args originally and copy/pasting had pulled it through).

PR #824, opened 2018-02-02
IE/Edge browser compatibility features

JackMorganNZ:
- Update details polyfill to 2.0.0
- Increase contrast of details elements
  This is done for two reasons:
  - In IE/Edge browsers currently, the border is not shown therefore the background of the detail element should stand out from the page body.
  - Increased contrast for those with disabilities.
- Fix for displaying images correcting in Internet Explorer
- Add CSS styling to display detail elements better on IE/Edge


JackMorganNZ:
This is what is looked like before PR (IE left, Edge right):

image

This is what is looked like after PR (IE left, Edge right):

image

PR #862, opened 2018-02-05
Only prepend www for production website (fixes #860)

### LINKED ISSUE #860: Only prepend www for production website
It should not be used for development website.

JackMorganNZ:


csunplugged/config/settings/production.py
```diff
@@ -23,7 +23,10 @@
 
 # URL Configuration
 # ------------------------------------------------------------------------------
-PREPEND_WWW = True
+if env("DEPLOYMENT", default="none") == "prod":  # noqa: F405
```
hayleyavw: is this intentionally a string rather than a type (`None`)

Issue #688, opened 2017-11-20
Implement versioning of resource implementations


jordangriffiths01:

JackMorganNZ:
Marking as urgent as these is now required due to multiple languages being used (see #1031).

hayleyavw:
@JackMorganNZ did you have any ideas on how this might be done?

JackMorganNZ:
@hayleyavw I do. Essentially the idea is that a resource can have multiple versions (probably as completely duplicated modules) so a resource can be redesigned/updated without ruining already translated lessons that use the lesson. For example the Job Badges (#1031) will need changes and would be good to not break the existing Spanish and German resources.

PR #1035, opened 2018-08-30
Issue/1031 - update job badges printable to be translator friendly

### LINKED ISSUE #1031: Modify Job Badges printable to use one string per badge
Currently the way the strings are broken for translation doesn't support all languages (for example: Māori has a word after the emphasis of the role).

hayleyavw:
And now uses pixel coordinates rather than svgs to draw the text.

Also I manually tested it with Māori to see what happens with longer text and it behaves just fine :) 

This doesn't address #688 - perhaps a conversation required before merging?

Fixes #1031


hayleyavw:
good spotting

hayleyavw:
LGTM 👍

Issue #985, opened 2018-08-06
Add button on topics page to link to classic topic list


JackMorganNZ:

caitlind:
Tim suggested a button to link to the translations on the old site as well.

PR #1249, opened 2019-09-30
Adds "more activities at classic website" button to topics page

caitlind:
Resolves [issue/985](https://github.com/uccser/cs-unplugged/issues/985)


caitlind:
Noticed something right after submitting this, I think I've left in a redundant div at lines 78-79 and 103-104. Someone with more HTML experience, should this be removed or is it serving a purpose?

courtneycb:
Also don't forget to put resolves or fixes before the issue tag in your description! :)

caitlind:
> I have a suggestion to reduce code duplication:
> 
> If you move the code for your button to an html file (say, `classic-unplugged-button.html`) under`templates/generic/` (make sure to load i18n and static in that file too) and then replace your button code in `index.html` with `{% include "generic/classic-unplugged-button.html" %}` :) I think it should work.

Sounds good! The english text is slightly different between the normal case and the case where no topics are available, but I think it will still make sense if I make them the same.

PR #1258, opened 2019-10-20
Deprecate 'dev' command for helper script.

### LINKED ISSUE #466: Make all csu commands work with dev prefix
Super low priority, but this is driving me insane.

`./csu start` is a command that can be used without the `dev` prefix, but in the event that I accidentally use attempt to use `./csu dev start` out of habit of using `dev`, I think that this should still work (same goes for all non-dev commands).

courtneycb:
This PR modifies the `csu` script such that `dev` commands can now be run without the `dev` prefix. However if commands are given with `dev` they will still work.

Reason for deprecation: 
Only developers use this script anyway and it is annoying if we get into the habit of using `dev` and then a command won't run because we accidentally included the dev prefix when it wasn't expecting it.

Resolves #466

### DIFF
--- a/csu
+++ b/csu
@@ -20,22 +20,13 @@ GREEN='\033[0;32m'
 NC='\033[0m' # No Color
 
 cmd_helps=()
-dev_cmd_helps=()
 
 defhelp() {
-  if [ "$1" = '-dev' ]; then
-    local command="${2?}"
-    local text="${3?}"
-    local help_str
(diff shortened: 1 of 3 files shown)


courtneycb:
> Looking good, but the docs should be updated to reflect the changes

Updated documentation. Added some commands that were not listed and followed the new-sentence-new-line rule for readability.

PR #1393, opened 2020-07-07
Order glossary terms alphabetically in viewed language.

### LINKED ISSUE #1290: Glossary page should list entries in alphabetical order for viewed language
For example, Māori lists these glossary entries in the following order:

- Hātepe
- Mati Tāhūrua
- Rapu ā-tāhūrua
- Waematihiko
- Mati takitaki (Check digit)
- Kōpeketanga
- ...

These should be in alphabetical for each language.

courtneycb:
Resolves #1290

### DIFF
--- a/csunplugged/topics/views.py
+++ b/csunplugged/topics/views.py
@@ -391,7 +391,7 @@ class GlossaryList(generic.ListView):
 
     def get_context_data(self):
         """Get context data for template rendering."""
-        term_locale = "term_" + get_language()
+        term_locale = "term_" + get_language().replace("-", "_")
         return {
             "glossary_terms": GlossaryTerm.objects.filter(
                 Q(languages__contains=[get_language()])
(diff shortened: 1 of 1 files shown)


csunplugged/topics/views.py
```diff
@@ -391,16 +391,16 @@ def get_queryset(self):
...
+        term_locale = "term_" + get_language()
         return {
             "glossary_terms": GlossaryTerm.objects.filter(
                 Q(languages__contains=[get_language()])
-            ).order_by("term_en"),
+            ).order_by(term_locale),
             "untranslated_glossary_terms": GlossaryTerm.objects.filter(
                 ~Q(languages__contains=[get_language()])
             ).order_by("term_en")
         }
```
eAlasdair, line 403:
```suggestion


```

eAlasdair:
image
The last three of these languages all break the glossary with Field Errors, but the latter two aren't available on the live csu :confused-parrot:

courtneycb:
> image
> The last three of these languages all break the glossary with Field Errors, but the latter two (translation modes) aren't available on the live csu :confused-parrot:

It's because the fields contain hyphens. Fixed.

PR #1397, opened 2020-07-13
Host videos on Vimeo instead of YouTube

### LINKED ISSUE #1181: Switch YouTube videos to Vimeo
YouTube now requires to show related videos after a video has finished, which is not useful for educational purposes. We should switch our videos to Vimeo, as we have done for our other projects.

courtneycb:
Resolves #1181. Note that this only replaces UCCSER owned YouTube videos.

Videos can be found under the 'CS Unplugged' folder on Vimeo.

### DIFF
--- a/subtitles/en/parity-magic-trick.vtt
+++ b/subtitles/en/parity-magic-trick.vtt
@@ -6,9 +6,8 @@ NOTE
 Computer Science Education Research,
 University of Canterbury, New Zealand
 Subtitle file for the video "CS Unplugged: Parity Magic"
-Author: Alasdair Smith
+Author: Alasdair Smith 13/02/2017
 Edited by: Courtney Bracefield 05/08/2020
-Date: 13/02/2017
 
 00:00.000 --> 00:04.000
 CS Unplugged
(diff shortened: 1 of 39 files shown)


courtneycb:
@eAlasdair ready for re-review

eAlasdair:
There's an option of subtitles now but they don't show on the site nor on Vimeo itself.

Nice work on the pause though

courtneycb:
> There's an option of subtitles now but they don't show on the site nor on Vimeo itself.
> Also would probably be best if subtitles were enabled by default - particularly on non-english versions even if the right language isn't available yet

I don't get this problem?
image
image

courtneycb:
> > I don't get this problem?
> 
> I still do, chrome & FF, as well as chrome on my laptop
> 
> Is it because you're logged into vimeo maybe?

Can you come in and show us in the office? Jack isn't getting the problem either.

subtitles/en/parity-magic-trick.vtt
```diff
@@ -11,7 +11,7 @@ Date: 13/02/2017
```
eAlasdair, line 11: Maybe say that you edited this file, though the translators didn't say anything either

PR #1405, opened 2020-08-04
Put testing examples for programming challenges into a separate markdown file

### LINKED ISSUE #1429: Solution provided for programming challenge on normal CSU site is not valid for same challenge on the plugging-it-in site.
We provide a python solution for [this programming challenge](https://csunplugged.org/en/topics/binary-numbers/programming/binary-notes-representing-number/) on the normal CSU site. However if we copy that solution into the IDE for the same question [on the plugging-it-in site](https://cs-unplugged-dev.appspot.com/en/plugging-it-in/binary-numbers/how-binary-digits-work/binary-notes-representing-number/) it isn't a valid solution. We need to decide whether we update the plugging-it-in site or the normal CSU site.

[Here](https://github.com/uccser/cs-unplugged/blob/develop/csunplugged/topics/content/en/binary-numbers/programming-challenges/binary-notes-representing-number/python-solution.md) is the code for the solution we provide.

image

We should check the other solutions as well. E.g the other binary notes question has the same error.

courtneycb:
Place testing examples into their own markdown file.
This was done using a python script.

Also ensures the solutions provided on the standard CSU site pass the tests for the respective programming challenge in the CSU plugging-it-in site.
Resolves #1429


csunplugged/topics/content/en/error-detection-and-correction/programming-challenges/check-valid-credit-card/testing-examples.md
```diff
@@ -0,0 +1,16 @@
+## Testing examples:
+
+Your program should display the outputs shown in this table for the given
+inputs provided:
+
+*Important: None of these is an actual credit card number, but the correct
```
eAlasdair, line 6:
```suggestion
*Important: None of these are actual credit card numbers, but the correct
```

eAlasdair: Might as well do a grammar check too

eAlasdair (commented):
Is there any code to _show_ the content of these new files? It's hard to test without that

csunplugged/topics/content/en/binary-numbers/programming-challenges/dots-largest-to-1/python-hints.md
```diff
@@ -1,7 +1,7 @@
 -   Make a variable called `number_of_dots` and set it to the largest number
     of dots entered as the input.
--   Divide the value of `number_of_dots` by 2 and display the result as the
-    output while `number_of_dots` is greater than or equal to 1.
+-   Divide the value of `number_of_dots` by 2, convert the result to an integer using `int()`
+    and display it as the output while `number_of_dots` is greater than or equal to 1.
```
eAlasdair, line 4: int() works here because it's definitely acting on halved powers of 2, but it's really bad practice in general. I can see users having two questions:
- *"If I halve the number 8 I get the number 4, so why do I need to convert it to an int?"*
- **After applying the solution to other situations**
*Why does 3.9 get rounded down to 3?*

So we should probably do one of:
- Don't say how to convert it, and just leave the user to figure it out themselves
- Explain that int() trims off the following decimal values, as well as why they're there
- Use // instead of / in the solution and hint, and explain what // does

courtneycb: I decided to use round()

eAlasdair (approved):
Now it might be confusing because you're rounding `n.0` to `n` and beginners still don't know why it's a float in the first place. But I suppose this is the most concise way of solving the problem

PR #1415, opened 2020-08-11
Use Djano's JavaScript translation catalog

### LINKED ISSUE #1409: Use Django's JavaScript Catalog
We should use Django's [JavaScript translation catalog](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#internationalization-in-javascript-code) to allow us to translate text in JavaScript files.

Refer to https://github.com/uccser/cs-field-guide/pull/713 and https://github.com/uccser/cs-field-guide/pull/731

courtneycb:
This PR enables the use of Django's [JavaScript translation catalog](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#internationalization-in-javascript-code), meaning we can now use gettext() to translate text in JavaScript files.

It also uses [django-statici18n](https://django-statici18n.readthedocs.io/en/latest/index.html), a Django app that collects JavaScript catalogs into a single location. (Instead of generating it dynamically on each request).

The docs have also been updated.

The following changes were also made in the `csu` helper script:

- Compiles messages within the `static` command as compiling JavaScript translations creates a static file.
- Remove deletion of volumes as this deletes a Docker volume creating for storing node modules. Tidying Docker should be a user choice/action, not a project action.

Resolves #1409


csu
```diff
@@ -311,14 +308,6 @@ cmd_clean() {
   echo "If the following commands return an argument not found error,"
   echo "this is because there is nothing to delete for clean up."
 
-  echo
-  echo "Deleting unused volumes..."
```
eAlasdair, line 315: > Remove deletion of volumes as this deletes a Docker volume creating for storing node modules. Tidying Docker should be a user choice/action, not a project action.

This is the clean command, which is exactly the 'user choice/action' you want removing volumes to be isn't it?

But more to the point, I don't understand the justification for removing this both here and in the end command, considering:
- We added the removal of volumes because they filled the allocated disk space of my PC to the point where I couldn't log into linux anymore
- It's supposed to be "Deleting **unused** volumes..."
- I assume this PR is based on how it works in CSFG, which has no problem (volumes are removed in the `./csfg end` command, but not the `clean` one surprisingly)

so
- Don't remove this from the clean command
- Please explain why it should be removed from the end command

courtneycb: Ah I see, I was following [Jack's PR](https://github.com/uccser/cs-field-guide/pull/731/files) without double checking :paddlin: I will add back in the code for deleting unused docker volumes in both the `end` and `clean` commands.

requirements/base.txt
```diff
@@ -36,6 +36,7 @@ django-modeltranslation==0.15.1
 uniseg==0.7.1
 python-bidi==0.4.2
 django-bidi-utils==1.0
+django-statici18n==1.8.2
```
eAlasdair, line 39: This should be licenced appropriately

https://github.com/zyegfryed/django-statici18n

Issue #889, opened 2018-02-19
Update Google Analytics


JackMorganNZ:
- Use the new `gtag.js` tracking.
- Track printable downloads.

courtneycb:
I have created an account on Google Tag Manager and added a container for CS Unplugged on there. 
Inside this container I have set up a tag to track PDF downloads using both GA and GTM, however I cannot test it until the GTM snippet has been added to the website. Will make a PR to add this snippet.

courtneycb:
Can we use Google Tag Manager instead?

courtneycb:
We should consider adding an [IP address filter](https://support.google.com/analytics/answer/1034840?hl=en) to exclude internal traffic.

courtneycb:
Google Tag Manager snippet added in #1435. Tags to track Printable downloads and Page views published in GTM and tested. You can view the version of GTM published with the admin account [here](https://tagmanager.google.com/?utm_source=marketingplatform.google.com&utm_medium=et&utm_campaign=marketingplatform.google.com%2Fabout%2Ftag-manager%2F#/versions/accounts/6001848934/containers/32212070/versions/1).

PR #1425, opened 2020-08-27
Use Google Tag Manager

courtneycb:
Adds the Google Tag Manager snippet to the website. This will enable us to use GTM to track pageviews and PDF downloads as I have set up the event tags for this in GTM already.
Note that the tags added to GTM cannot be tested until this snippet is added to the website. Therefore if it doesn't work there will be no pageview tracking as this replaces the old Google Analytics code for that.

Relates to #889 

**Do not merge this until the survey on develop has been completed.**


eAlasdair (approved):
There's not much I can do to test this

PR #1434, opened 2020-09-08
Try implementing Django's JavaScript translation catalog again

### LINKED ISSUE #1409: Use Django's JavaScript Catalog

courtneycb:
In #1415 an attempt was made at using Django's JavaScript translation catalog. However an error occurred (detailed in #1409) and the changes had to be rolled back due to a survey running on the develop site at the time.

This PR implements the same changes that were in #1415 but also has the changes to create the `djangojs` message file.

Hopefully resolves #1409


eAlasdair (approved):
It looks fine but I also missed the first bug -_-

Think about whether or not this should go into the current release though

courtneycb:
> It looks fine but I also missed the first bug -_-
> 
> Think about whether or not this should go into the current release though

Not putting this into the current release

eAlasdair:
Since this also adds the gettexts, this PR should implement them into everything added in the current release once it's out

PR #1444, opened 2020-09-23
Match footer colour to header colour

courtneycb:
We should keep the footer colour consistent with the header. Done for plugging it in and at home.

Old:
image

New:
image (6)


eAlasdair (approved):
Looks good

PR #1448, opened 2020-09-29
Add walk-through for plugging it in

courtneycb:
Add step by step explanations for each main component of the programming challenge page. This is triggered by a button located underneath the test results table. 

This PR is a result of feedback from Dana's survey.


csunplugged/static/js/plugging-it-in-walkthrough.js
```diff
@@ -0,0 +1,103 @@
...
+            },
+            {
+                element: "#next_challenge_button button",
+                intro: "Click this button when you are ready to move on to the next challenge.",
+                step: 11,
+                position: "top"
+            },
+            {
+                element: "#lessons_nav_toggle",
+                intro: "You can navigate to other programming challenges here.",
```
eAlasdair, line 67: Maybe be consistent with the "Click [this] to do [this]" type language for the other buttons

Like:
Click here to see the other challenges for this topic

eAlasdair: On that note it's not immediately obvious how to get from a question back to the topic selector, only to questions within the topic. It's the unplugged homepage (top left button) but peeps might not recognise that

eAlasdair (changes requested):
Good work! Also on the last question the next question button doesn't exist, so it highlights the top left corner

eAlasdair (approved):
Looking good! We should consider getting an author to look over the language used in the tips to ensure it's right for our target audience

PR #1449, opened 2020-09-30
Improve consistency of blues in the PII colour scheme

eAlasdair:
Reduces the number of different shades of blues used in the Plugging It In area of the website:
- Navigation buttons, like `More information for educators` are now the main theme colour
- The `X` and horizontal line in the challenge hamburger menu is now the main theme colour
- The `How does this page work?` button on the challenge page now uses the same hover colour as the buttons `Check` and `Download`

Note that I renamed the file `plugging-it-in-footer.scss`, as somewhere to put the new `pii-navigation` class. I'm happy to rename or relocate the class if there's a better place for it

I also replaced several re-definitions of preset colours with their variable names

### DIFF
--- a/csunplugged/static/scss/plugging-it-in-base.scss
+++ b/csunplugged/static/scss/plugging-it-in-base.scss
@@ -1,0 +1,19 @@
+@import "core-variables";
+
+#page-footer {
+    background-color: $plugging-it-in-colour;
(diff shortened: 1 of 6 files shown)


courtneycb (approved):
Nice work!

PR #1452, opened 2020-10-04
Fix PII edge case where neither Next Challenge nor Previous Challenge buttons are needed, AND fix bug where translation not available message was not displayed

eAlasdair:
If there is only one challenge, previously the 'Next challenge' button on the challenge page would be hidden, but not the 'Previous challenge' one.

Now both are properly hidden


courtneycb (approved):

eAlasdair:
Accidentally fixed a different bug in this branch:

Fixes bug where if no translation is available, the challenge attempt page would be blank:
image

- Adds 'homepage' variable to the generic translation error template
- Directs user either to the homepage (PII home) or the topic page (Challenge page)

courtneycb (changes requested):
Nice :+1:

courtneycb (approved):

PR #1453, opened 2020-10-04
Add link to PII from programming challenges on educators site

courtneycb:
Instead of showing educators what their program should output, we should encourage them to use our new Plugging It In system to develop and run their program for the challenge. These pages are eventually going to be replaced by the Plugging It In system. This change only applies to the Python section as Scratch programs are not yet supported in PII.

Note:
- The link to the challenge in PII requires a lesson slug. The DB currently allows a many-to-many relationship between programming challenges and lessons. This means it could be possible that one programming challenge is associated with multiple lessons. However, there are currently no programming challenges associated with more than one lesson and since this is a temporary addition to the site, we provide the 'first' lesson to be used as the lesson in the link.

- The 'Hints' section is provided on the PII challenge so is removed from the original Python challenge box.

Old:
image

New:
image


csunplugged/static/scss/website.scss
```diff
@@ -863,3 +863,21 @@ button:hover .icon-search {
         opacity: 0.5;
     }
 }
+
+.pii-navigation {
```
eAlasdair, line 867: Could you change pii here and below to `plugging-it-in` for consistency with a similar one further up?

Also maybe subclass this as `.btn.plugging-it-in-navigation` (if that's correct syntax) to make it clear this is for buttons

eAlasdair (changes requested):
looking good

eAlasdair (approved):

PR #1454, opened 2020-10-06
Prevent navbar from cutting off content on smaller screens for PII

courtneycb:
Old:
image (8)

New:
image


eAlasdair:
Is it possible to do the same thing to the homepage?
image

eAlasdair:
> Is it possible to do the same thing to the homepage?

The homepage and programming challenge page both use the `base-plugging-it-in.html` file, so applying this fix there _instead_ of where it is would be better

eAlasdair (approved):

eAlasdair:
Give it another check before you merge

eAlasdair (approved):

Issue #1455, opened 2020-10-06
PII walkthrough partially unnavigable on small screens


eAlasdair:
On screens smaller than the bootstrap defined `md` size this happens at the description of the 'Next Challenge' button:

![image](https://user-images.githubusercontent.com/25916475/95271241-dfa3bb00-0899-11eb-92c6-7f3b7a88f685.png)

At its worst, the guide box doesn't appear anywhere on the screen, making it impossible to navigate the walkthrough.

PR #1456, opened 2020-10-06
Hide pii walkthrough on small screens

eAlasdair:
Relates to #1455 

Draft because we should answer the question: `Do we want to 'fix' it this way?` before we think about actually merging it


courtneycb:
I think we do this in [CodeWOF](https://github.com/uccser/codewof/pull/304/files#diff-9ae49ebe49e8a6afc97221955221f23bR319)?

courtneycb (approved):

PR #1460, opened 2020-10-12
Encode Plugging it in challenge user answers

eAlasdair:
If a user (intended or otherwise) enters HTML code into the answer box that is subsequently saved, then when the user reloads the page an error occurs that prevents most elements on the page from being used until the user deletes their cookies.

This PR adds a module `Scramble.js` that uses the Vigenere cipher to encode a submitted answer such that when it is reloaded, the chance of it triggering that error becomes negligible.
I chose Vigenere purely because I'm currently studying it as part of a cryptography course

*The 16-character cipher key is created at the time the message is encoded, and saved as plaintext in the database*
**As such, this module should not be used for security, nor considered secure in any way**
The point is to make it not look like HTML, rather than have it be encrypted.

This will work because:
- Since the key is made of alphanumeric characters, each character in the encoded string is no less than 48 characters further down the UTF-16 list than it's plaintext equivalent
- We don't view user submissions, so there's no need to store them in plaintext. But also the key is stored with the encoding so it's also easily decoded

The module also has a version number: if the program is changed then anything encoded prior to that can't be decoded, so this number is used to track which version was used to encode the string.


JackMorganNZ:
I seem to be missing something because this seems overly complicated, and I don't understand what step is causing it to break.

Would [escaping](https://docs.djangoproject.com/en/3.1/ref/utils/#module-django.utils.html) the submitted string work as a simpler alternative?

eAlasdair:
> I seem to be missing something because this seems overly complicated, and I don't understand what step is causing it to break.
> 
> Would [escaping](https://docs.djangoproject.com/en/3.1/ref/utils/#module-django.utils.html) the submitted string work as a simpler alternative?

This is definitely overly complicated, I got too focused on implementing a thing I'm currently studying and didn't give enough time to think about other options, sorry.


I've since realised a better solution could be to avoid loading the previous submissions in the HTML template in the first place, instead just request them in Javascript.
Alternatively, yes, escaping the string could also work.

PR #1464, opened 2020-10-14
Hotfix 6.0.1: Make previous submission for a PII challenge load straight into the answer box

### LINKED ISSUE #1463: HTML can be inserted into a user's Plugging it in challenge templates
### The steps are:
- [Open any programming challenge on PII](https://www.csunplugged.org/en/plugging-it-in/binary-numbers/how-binary-digits-work/binary-numbers-no-calculations/)
- Submit this code in the answer box:
    - `</script><script type="text/javascript">console.log("hi")</script>`
        - It must be all on one line
        - Adapt to your desired level of maliciousness
- Refresh the page

When the user re-requests the page, the submitted 'code' is loaded [in the HTML template](https://github.com/uccser/cs-unplugged/blob/develop/csunplugged/templates/plugging_it_in/programming-challenge.html#L191). The first `</script>` closes the script that assigns the variables, and what follows is loaded as a new HTML script tag. The resulting HTML is sent to the user and executes 'as normal'.

I haven't been able to submit code that executes while also not breaking every PII challenge page. When the page does break, deleting your cookies for CSU allows you to get it working again

### Three potential fixes
- [Escape the submitted code.](https://docs.djangoproject.com/en/3.1/ref/utils/#module-django.utils.html)
    - It was already escaped, the problem is that [it gets un-escaped](https://docs.djangoproject.com/en/2.2/ref/templates/language/#automatic-html-escaping) in order to load as a variable.
        - So we would need to un-escape it in JS rather than HTML, and doing so will be tricky
(`&#39;}}&lt;` etc is loaded into the answer box)
- Don't load the user submitted code into the HTML template, and instead request it in JS
    - Will be slower, another request is needed
- Load the submitted code in the HTML, not as a variable, but straight into the <textarea> answer box.
    - This is how it's done in CodeWOF
    - Since it's in a <textarea> everything is rendered as plaintext

eAlasdair:
Resolves #1463 

This method mimics [the way we do it in CodeWOF](https://github.com/uccser/codewof/blob/develop/codewof/templates/programming/question_components/editor-python.html). It avoids the JS step by just putting it in the answer box to begin with


eAlasdair:
Will reformat as a hotfix PR

courtneycb (dismissed):

courtneycb (approved):

PR #2070, opened 2022-12-14
Add deletion checks to even more loaders

Yamboy1:
I missed a few deletion checks on loaders


Yamboy1:
It turns out that at_home doesn't have any tests on it whatsoever, and I would rather merge this pull request with at least some tests, so this will stay open until tests get written for it.

